PR #14, opened 2021-11-09
Wrap ownerOf function in try/catch:

atbe:


src/modules/nft.ts
```diff
@@ -73,8 +73,19 @@ export class NFTModule extends ModuleWithRoles<NFT> {
...
+   * @param tokenId - ID of the NFT to get the owner of
+   * @returns the owner of the token, or an empty string if the
+   * token has been burned
+   */
   public async ownerOf(tokenId: string): Promise<string> {
-    return await this.readOnlyContract.ownerOf(tokenId);
+    try {
+      return await this.readOnlyContract.ownerOf(tokenId);
+    } catch (e) {
+      return "";
```
jakeloo, line 87: do `ethers.constants.AddressZero` for whenever we return addresses.

src/modules/nft.ts
```diff
@@ -73,8 +73,19 @@ export class NFTModule extends ModuleWithRoles<NFT> {
...
+   *
+   * @param tokenId - ID of the NFT to get the owner of
+   * @returns the owner of the token, or an empty string if the
+   * token has been burned
+   */
   public async ownerOf(tokenId: string): Promise<string> {
-    return await this.readOnlyContract.ownerOf(tokenId);
+    try {
+      return await this.readOnlyContract.ownerOf(tokenId);
+    } catch (e) {
```
atbe, line 86: I should call out that this is a pretty general error handler here and that we may signal that a token is burned even if the call failed for a different reason.

jakeloo (changes requested):

src/modules/nft.ts
atbe: Good catch, fixed!

atbe:
Please squash merge!

jakeloo:
feel free to merge it and make a patch build.

@jnsdls the process of making the sdk build is:
1. merge into main
2. yarn build
3. np

right?

jnsdls:
> feel free to merge it and make a patch build.
> 
> @jnsdls the process of making the sdk build is:
> 
> 1. merge into main
> 2. yarn build
> 3. np
> 
> right?

yes

PR #19, opened 2021-11-10
Royalty module

atbe:


src/modules/royalty.ts
```diff
@@ -0,0 +1,152 @@
+import { Royalty, Royalty__factory } from "@3rdweb/contracts";
+import { BigNumberish } from "ethers";
+import { ModuleType, Role, RolesMap } from "../common";
+import { Currency, getCurrencyMetadata } from "../common/currency";
+import { Module } from "../core/module";
+
+interface IRoyalty {
```
atbe, line 7: @jakeloo @jnsdls please let me know what you guys think about this interface before I fill out the methods.

jakeloo: is this public interface?

src/modules/royalty.ts
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * Get the total shares held by payees
+   */
+  totalShares(): Promise<BigNumberish>;
```
jakeloo, line 11: i know this is going to be inconsistent but i think we should:
ALL input of big number type / id = BigNumberish
ALL output / result of big number = BigNumber

reason being that on the input side we'd support number, string or big number.
output = it's expected that it's always bignumber and not int / string

src/modules/royalty.ts
```diff
@@ -0,0 +1,152 @@
...
+   * @param tokenAddress - The ERC20 token address to check royalty distributions for
+   */
+  releasedByToken(address: string, tokenAddress: string): Promise<BigNumberish>;
+
+  /**
+   * Gets the address of the payee at index `index`
+   *
+   * @param index - The index of the payee to get the address of
+   */
+  payee(index: BigNumberish): Promise<string>;
```
jakeloo, line 53: i think we probably want something like `getAll` that returns a list of people involves in this maybe including their respective shares.

jakeloo: not sure if `payee` is the right name for this too.

src/modules/royalty.ts
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ *
+ * Access this module by calling {@link ThirdwebSDK.getRoyaltyModule}
+ * @public
+ */
+export class RoyaltyModule extends Module implements IRoyalty {
+  public static moduleType: ModuleType = ModuleType.ROYALTY as const;
+
+  public static roles = [
```
jnsdls, line 81: it has roles but does not extend ModuleWithRoles?

src/modules/royalty.ts
atbe: I'll make it public

src/common/module-type.ts
```diff
@@ -14,6 +14,7 @@ export enum ModuleType {
   MARKET = 6,
   DROP = 7,
   DATASTORE = 8,
+  ROYALTY = 9,
```
jnsdls, line 17: I think this has to line up with something in the contracts? But I might be misremembering @jakeloo can u confirm?

src/modules/royalty.ts
```diff
@@ -0,0 +1,152 @@
...
+   */
+  payee(index: BigNumberish): Promise<string>;
+
+  /**
+   * Triggers the release of the amount of tokens owed to `address` according
+   * to their percentage of the total shares and their previous withdrawals.
+   *
+   * @param address - The address of the payee to release the tokens for
+   */
+  release(address: string): Promise<void>;
```
jakeloo, line 61: i think currently, most write methods return TransactionReceipt if void.
for a reason that people can keep track of the hashes offchain or do further processing.

in general, i think we need to have a better system in the SDK for TransactionReceipt, i feel like we should almost have a `.on('transactionSent', (transactionReceipt) => void)` event in the SDK so people can subscribe for these tx receipt. (for future notes)

src/modules/royalty.ts
```diff
@@ -0,0 +1,152 @@
...
+    RolesMap.minter,
+    RolesMap.pauser,
+    RolesMap.transfer,
+  ] as const;
+
+  /**
+   * @override
+   * @internal
+   */
+  protected getModuleRoles(): readonly Role[] {
```
jnsdls, line 92: same question, RE role interfaces when it does not extend a module with roles?

src/common/module-type.ts
jakeloo: nope, the contract is now agnostic. any integer you pass in is going to be valid. :)

src/modules/royalty.ts
jakeloo: we can remove roles methods since it doesn't extend roles

src/common/module-type.ts
jnsdls: sheesh ok

src/modules/app.ts
```diff
@@ -194,6 +198,21 @@ export class AppModule extends Module<ProtocolControl> {
     }));
   }
 
+  /**
+   * Method to get a list of Royalty module metadata.
+   * @alpha
+   * @returns A promise of an array of Royalty modules.
+   * @deprecated - Use {@link AppModule.getAllModuleMetadata} instead
```
jakeloo, line 205: deprecated heh

src/modules/app.ts
```diff
@@ -194,6 +198,21 @@ export class AppModule extends Module<ProtocolControl> {
...
+   */
+  public async getRoyaltyModules(): Promise<ModuleMetadata[]> {
+    return (
+      await this.getAllContractMetadata(await this.getDatastoreAddress())
```
jakeloo, line 209: Datastore heh

src/modules/royalty.ts
jnsdls: this makes sense to me and I like the idea of this being the case always

src/modules/royalty.ts
```diff
@@ -0,0 +1,152 @@
...
+export interface IRoyaltyModule {
+  /**
+   * Get the total shares held by payees
+   */
+  totalShares(): Promise<BigNumberish>;
+
+  /**
+   * Get the total amount of royalty already released
+   */
+  totalReleased(): Promise<BigNumberish>;
```
jnsdls: the whole "release" stuff makes no sense to me as someone that hasn't dove deep into royalty specific functionality. I'm super confused why I would call this method

src/modules/royalty.ts
```diff
@@ -0,0 +1,152 @@
...
+   * @param tokenAddress - The ERC20 token address to check royalty distributions for
+   */
+  totalReleasedByToken(tokenAddress: string): Promise<BigNumberish>;
+
+  /**
+   * Gets the amount of shares held by an account at the address `address`
+   *
+   * @param address - The address to check the shares held for
+   */
+  shares(address: string): Promise<BigNumberish>;
```
jnsdls: what are shares? Am I investing in a royalty? Can I buy them on a secondary market?

jnsdls:
maybe I'm just too dumb to understand the concept of a royalty module, but none of the comments help me understand why I would call them

src/core/index.ts
```diff
@@ -385,6 +387,15 @@ export class ThirdwebSDK {
     return this.getOrCreateModule(address, DropModule);
   }
 
+  /**
+   *
+   * @param address - The contract address of the given Royalty module.
+   * @returns The Drop Module.
```
jnsdls, line 393: the splits module

src/modules/royalty.ts
```diff
@@ -0,0 +1,203 @@
...
+   */
+  getRecipientSplitPercentage(address: string): Promise<SplitRecipient>;
+
+  /**
+   * Returns the amount of royalty available for a recipient
+   * to withdraw in the native currency.
+   *
+   * @param address - The address of the recipient to check the balance of.
+   */
+  balanceOf(address: string): Promise<BigNumber>;
```
jnsdls, line 34: not sure if I want an additional `balance` method as well as `balanceOf` (balance would just get *my* balance if I'm connected, but the more I think about it the more I think we should instead get rid of `balance` everywhere else and force users to explicitly pass the address....)

src/modules/royalty.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  /**
+   * Returns the amount of royalty available for a recipient
+   * to withdraw in the native currency in a specific currency.
+   *
+   * @param walletAddress - The address of the recipient to check the balance of.
+   * @param tokenAddress - The address of the currency to check the balance in.
+   */
+  balanceOfByToken(
```
jnsdls, line 43: this name is annoying but I understand it is necessary :D

jnsdls: maybe `erc20BalanceOf()` instead actually? Just thinking out loud

src/modules/royalty.ts
```diff
@@ -0,0 +1,203 @@
...
+   * the `address` is owed and transfer it to the wallet, in the
+   * currency specified by `tokenAddress`.
+   *
+   * For example: If the native currency of a chain is ETH but the user
+   * wants to withdraw their split in $MATIC, they should pass
+   * the address of the $MATIC token as the `tokenAddress` parameter.
+   *
+   * @param walletAddress - The address to withdraw royalties for.
+   */
+  withdrawToken(walletAddress: string, tokenAddress: string): Promise<void>;
```
jnsdls, line 67: `withdrawErc20` (if we change the balance method)?

src/modules/royalty.ts
```diff
@@ -0,0 +1,203 @@
...
+  public async get(): Promise<Currency> {
+    return await getCurrencyMetadata(this.providerOrSigner, this.address);
+  }
+
+  public async getAllRecipients(): Promise<SplitRecipient[]> {
+    const recipients: SplitRecipient[] = [];
+
+    let index = BigNumber.from(0);
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
```
jnsdls, line 101: while loops scare me but I trust u

src/modules/royalty.ts
```diff
@@ -0,0 +1,203 @@
...
+      try {
+        const recipientAddress = await this.readOnlyContract.payee(index);
+        recipients.push(
+          await this.getRecipientSplitPercentage(recipientAddress),
+        );
+        index = index.add(1);
+      } catch (err: any) {
+        // The only way we know how to detect that we've found all recipients
```
jnsdls, line 109: is this a contract limitation that the contract could address? @jakeloo? this *seems* bad just generally

src/modules/royalty.ts
```diff
@@ -0,0 +1,203 @@
...
+      ),
+    );
+    return await getCurrencyValue(this.providerOrSigner, tokenAddress, value);
+  }
+
+  public async withdraw(address: string): Promise<void> {
+    await this.sendTransaction("release(address)", [address]);
+  }
+
+  private async _pendingPayment(
```
jnsdls, line 180: I would *like* it if private & public methods were not mixed together ... but this is truly just personal preference

src/types/SplitRecipient.ts
```diff
@@ -0,0 +1,14 @@
+export interface SplitRecipient {
```
jnsdls, line 1: should probably also document the top level recipient interface itself, not crucial though

src/modules/royalty.ts
jakeloo: i think token might be more aligned with our sdk because we now call erc20 tokens token

src/modules/royalty.ts
```diff
@@ -0,0 +1,203 @@
...
+  ): Promise<SplitRecipient> {
+    const [totalShares, walletsShares] = await Promise.all([
+      this.readOnlyContract.totalShares(),
+      this.readOnlyContract.shares(address),
+    ]);
+
+    return {
+      address,
+      splitPercentage:
+        walletsShares.mul(BigNumber.from(1e7)).div(totalShares).toNumber() /
```
jakeloo, line 136: we might want to make it clear that 1e7 / 1e9 = for decimal precision

PR #24, opened 2021-11-13
Module deployments

atbe:


test/app.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      name: "Testing module from SDK",
+      sellerFeeBasisPoints: 1000,
+      image:
+        "https://pbs.twimg.com/profile_images/1433508973215367176/XBCfBn3g_400x400.jpg",
+    });
+
+    console.log("DEPLOYED MODULE =", module);
+  });
+
+  it.skip("should deploy a splits module successfully", async () => {
```
atbe, line 90: @jakeloo @jnsdls 

So its important that we use `shares` here and not %'ages

I tried to communicate this as best as I could in the `NewSplitRecipient.ts` file, but it might be a bit confusing.

Let's just make sure we highlight very clearly on the guide and people expect that `recipient percentage = recipient_shares / sum(all shares)`

atbe:
This PR is ready for review @jakeloo @jnsdls 

I'm publishing another dev package now

src/interfaces/IAppModule.ts
```diff
@@ -0,0 +1,42 @@
...
+import CurrencyModuleMetadata from "../types/module-deployments/CurrencyModuleMetadata";
+import DatastoreModuleMetadata from "../types/module-deployments/DatastoreModuleMetadata";
+import DropModuleMetadata from "../types/module-deployments/DropModuleMetadata";
+import MarketModuleMetadata from "../types/module-deployments/MarketModuleMetadata";
+import NftModuleMetadata from "../types/module-deployments/NftModuleMetadata";
+import PackModuleMetadata from "../types/module-deployments/PackModuleMetadata";
+import SplitsModuleMetadata from "../types/module-deployments/SplitsModuleMetadata";
+
+export default interface IAppModule {
+  deployCollectionModule(
```
jakeloo, line 19: Bundle

src/interfaces/IAppModule.ts
```diff
@@ -0,0 +1,42 @@
...
+import CollectionModuleMetadata from "../types/module-deployments/CollectionModuleMetadata";
+import CurrencyModuleMetadata from "../types/module-deployments/CurrencyModuleMetadata";
+import DatastoreModuleMetadata from "../types/module-deployments/DatastoreModuleMetadata";
+import DropModuleMetadata from "../types/module-deployments/DropModuleMetadata";
+import MarketModuleMetadata from "../types/module-deployments/MarketModuleMetadata";
+import NftModuleMetadata from "../types/module-deployments/NftModuleMetadata";
+import PackModuleMetadata from "../types/module-deployments/PackModuleMetadata";
+import SplitsModuleMetadata from "../types/module-deployments/SplitsModuleMetadata";
+
+export default interface IAppModule {
```
jakeloo, line 18: Do we need `deployXModule` or `deployX` is good enough?

src/interfaces/IAppModule.ts
```diff
@@ -0,0 +1,42 @@
...
+    metadata: CurrencyModuleMetadata,
+  ): Promise<CurrencyModule>;
+
+  deployMarketplaceModule(
+    metadata: MarketModuleMetadata,
+  ): Promise<MarketModule>;
+
+  deployPackModule(metadata: PackModuleMetadata): Promise<PackModule>;
+
+  deployDropModule(metadata: DropModuleMetadata): Promise<DropModule>;
```
jakeloo, line 37: is it Drops or Drop?

I think in general we have a little bit of inconsistent between drop vs drops, split vs splits. We should settle on one

src/interfaces/IAppModule.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  deploySplitsModule(metadata: SplitsModuleMetadata): Promise<SplitsModule>;
+
+  deployNftModule(metadata: NftModuleMetadata): Promise<NFTModule>;
+
+  deployCurrencyModule(
+    metadata: CurrencyModuleMetadata,
+  ): Promise<CurrencyModule>;
+
+  deployMarketplaceModule(
```
jakeloo, line 31: Market or marketplace? I think this should match `getX`

src/types/module-deployments/DropModuleMetadata.ts
```diff
@@ -0,0 +1,50 @@
...
+   * The symbol for the Drop Collection
+   */
+  @JsonProperty("symbol", String, PropertyConvertingMode.IGNORE_NULLABLE)
+  symbol?: string;
+
+  /**
+   * The max supply (required)
+   */
+  @JsonProperty("max_supply", Number, PropertyConvertingMode.IGNORE_NULLABLE)
+  maxSupply = 1;
```
jakeloo, line 35: I think we probably want to default to 0 or if it's required, we probably don't need a default value because its kinda confusing

src/types/module-deployments/DropModuleMetadata.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * The IPFS base URI thats prepended to all token URIs. This allows
+   * you to batch upload all the metadata for the tokens ahead of time
+   * so they can be lazy minted later
+   *
+   * e.g. if a Drop Collection has a token URI of /ipfs/BOREDAPES/TOKEN1, then the IPFS base URI is /BOREDAPES
```
jakeloo, line 42: fyi: the contract only append `{token id}`, so technically the base uri is `/BOREDAPES/` :)

src/types/module-deployments/NewSplitRecipient.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface NewSplitRecipient {
+  /**
+   * The address of the recipient
+   */
+  address: string;
+
+  /**
+   * The number of shares for the recipient
+   */
+  shares: BigNumber;
```
jakeloo, line 25: we'd probably want BigNumberish for input data type so people can pass in string, number or big number

jakeloo: unless this is an output result type?

src/types/module-deployments/NewSplitRecipient.ts
```diff
@@ -0,0 +1,26 @@
...
+ * If there are two recipients each with 10 shares, they each will receive 50%
+ * of the total royalties.
+ *
+ * If there are two recipients each with 1 share, they each will receive 50%
+ * of the total royalties.
+ *
+ * I.e. the total number of shares is used to calculate the percentage of the
+ * total royalties that is allocated to each recipient.
+ */
+export interface NewSplitRecipient {
```
jakeloo, line 16: does the shares mechanism made sense for developers? if you're a new dev using this. would you understand what's the shares and how to split them based of total number of shares? (just wondering)

jakeloo: does it make sense to be `NewSplitRecipient` or `SplitRecipient` is fine

yarn.lock
```diff
@@ -3846,19 +3853,19 @@ forever-agent@~0.6.1:
   resolved "https://registry.yarnpkg.com/forever-agent/-/forever-agent-0.6.1.tgz#fbc71f0c41adeb37f96c577ad1ed42d8fdacca91"
   integrity sha1-+8cfDEGt6zf5bFd60e1C2P2sypE=
 
-form-data@^3.0.0:
-  version "3.0.1"
-  resolved "https://registry.yarnpkg.com/form-data/-/form-data-3.0.1.tgz#ebd53791b78356a99af9a300d4282c4d5eb9755f"
-  integrity sha512-RHkBKtLWUVwd7SqRIvCZMEvAMoGUp0XU+seQiZejj0COz3RI3hWP4sCv3gZWWLjJTd7rGwcsF5eKZGii0r/hbg==
+form-data@*, form-data@^4.0.0:
```
jakeloo, line 3856: we'll need to make sure that this doesn't break web + node build.
iirc last time it'd break web js build OR node js build.

src/interfaces/IAppModule.ts
atbe: I personally feel like `deployXModule` is clearer until we decide to remove the keyword `Module` completely

src/interfaces/IAppModule.ts
atbe: Nice catch, updating

src/interfaces/IAppModule.ts
atbe: `Drop` makes sense, `Splits` makes sense, `Drops` doesn't make sense to me, its an *NFT Drop*, not an *NFT Drops*

src/interfaces/IAppModule.ts
atbe: fair, this should be `Market` everywhere then

src/types/module-deployments/DropModuleMetadata.ts
atbe: It's just to get the type checking, it doesn't actually do anything. I need a value though so its initialized to something.

src/types/module-deployments/NewSplitRecipient.ts
atbe: Good call, this is input so I'll fix that

src/types/module-deployments/NewSplitRecipient.ts
atbe: I would assume so. Its not all that complicated and if we have good docs/tutorials, it should click very quickly. Otherwise the other option is using percentages which is its own mess that is not as clean imo.

src/interfaces/IAppModule.ts
jakeloo: i think there's a question on the product level. our sdk should match the branding of what we sell to our customers.
thirdweb.com and thirdweb.com/dashboard, we've branded it as `Drops` rather than `Drop`.

yarn.lock
atbe: We have a test package we can test in console so we should test it out. Building is all good.

src/types/module-deployments/DropModuleMetadata.ts
jakeloo: in general i think the type should really be a BigNumberish, but i'm not sure how that works with json2ts.
i think if that's the case we should default it to 0, i think `1` feels arbitrary.

src/interfaces/IAppModule.ts
jakeloo: Actually I think we're using DropModule in somewhere else, i think we should be fine. as long as we keep it consistent lol

src/types/module-deployments/DropModuleMetadata.ts
atbe: I can write a converter for json2ts, its simple and def something we want for the rest of the SDK.

src/common/ipfs.ts
```diff
@@ -1,6 +1,8 @@
-import FormData from "form-data";
 import { MetadataURIOrObject } from "../core/types";
 
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const FormData = require("form-data");
```
jnsdls, line 4: can we do this with a switch on

```ts
if(!globalThis.FormData){
  globalThis.FormData = require("form-data");
}
```

src/types/module-deployments/DropModuleMetadata.ts
atbe: I love how thorough you are, thanks Jake, fixed.

src/interfaces/IAppModule.ts
jnsdls: make it all singular.

jnsdls: everything is singular, no more plurals unless it's explicitly an array then *maybe* a plural

src/common/ipfs.ts
jnsdls: reason: this will otherwise break some web builds (because "form-data" lib includes node internal packages)

src/types/module-deployments/BundleModuleMetadata.ts
```diff
@@ -0,0 +1,31 @@
...
+export default class BundleModuleMetadata extends CommonModuleMetadata {
+  /**
+   * The amount of royalty collected on all royalties represented as basis points.
+   * The default is 0 (no royalties).
+   *
+   * 1 basis point = 0.01%
+   *
+   * For example: if this value is 100, then the royalty is 1% of the total sales.
+   */
+  @JsonProperty(
```
jnsdls, line 18: what is this black magic 🪄

src/types/module-deployments/NewSplitRecipient.ts
jnsdls: wouldn't this just be `SplitRecipient`? the "new" is obvious if you're *adding* it, no? does the `get` on this return an `OldSplitRecipient` interface?

jnsdls (dismissed):

src/common/ipfs.ts
jnsdls: @atbe this is really the only real blocker from my side

src/types/module-deployments/BundleModuleMetadata.ts
atbe: lol just handles json serialization/deserialization with support for mapping the serialized data to specific property names that might be different from the class names.

It's nice because to the caller they see `externalLink` but when we serialize to json, it'll be `external_link` (which is what opensea requires)

src/common/ipfs.ts
atbe: Hmm, so should I be replacing this line with: ??

```js
if(!globalThis.FormData){
  globalThis.FormData = require("form-data");
}
```
I'm not familiar with the `globalThis` keyword (googling it now)

atbe: Oh I see, so if this is true, we're running in nodejs because otherwise `FormData` would be defined on the `globalThis` object

src/types/module-deployments/NewSplitRecipient.ts
atbe: Its different. The `NewSplitRecipient` takes in an integer shares value, and the `SplitRecipient` returns back the % value we could make them both percentages, but then they have to ensure what they create always adds up to 100%

src/common/ipfs.ts
atbe: Alright I just tested this on console, all is well 👍

jnsdls: yep exactly

PR #37, opened 2021-11-19
Upgraded @3rdweb/contracts package, added two new properties to the d…

atbe:
…rop module metadata: primarySaleRecipientAddress and primarySaleFeeBasisPoints

Plan is to release a minor version of the sdk with this change and then plan for console to integrate it into the drop creation flow (cc @jnsdls)

### DIFF
--- a/src/modules/app.ts
+++ b/src/modules/app.ts
@@ -582,6 +582,10 @@ export class AppModule
     metadata: DropModuleMetadata,
   ): Promise<DropModule> {
     invariant(metadata.maxSupply !== undefined, "Max supply must be specified");
+    invariant(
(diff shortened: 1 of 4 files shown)


src/modules/app.ts
```diff
@@ -582,6 +582,10 @@ export class AppModule
     metadata: DropModuleMetadata,
   ): Promise<DropModule> {
     invariant(metadata.maxSupply !== undefined, "Max supply must be specified");
+    invariant(
+      metadata.primarySaleRecipientAddress !== "",
```
jakeloo, line 586: https://docs.ethers.io/v5/api/utils/address/#utils-isAddress

PR #38, opened 2021-11-19
[WIP] Drops interface - STILL IN REVIEW

atbe:


src/enums/ClaimConditionType.ts
```diff
@@ -0,0 +1,27 @@
+enum ClaimConditionType {
+  StartTime,
+  LimitPerWallet,
+  MerkleRoot,
+  WaitTimeBetweenClaims,
+}
+
+/**
+ * Questions:
```
atbe, line 9: @jakeloo @nkrishang need your input on all these questions, please. I'm trying to craft a `ClaimConditionsFactory`, but I need to understand the fundamental rules around all of our supported claim conditions first.

src/enums/ClaimConditionType.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * 1. Can there only be a single `startTimestamp` condition? What is a `startTimestamp` condition?
```
jakeloo, line 11: startTimestamp = seconds into the future when the mint condition will be active.

startTimestamp = 0 -> the condition will be active NOW.
startTimestamp = 60 -> the condition will be active in 60 seconds.

src/enums/ClaimConditionType.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * 2. What is `maxMintSupply`?
```
jakeloo, line 13: maximum number of mintable nft per condition

src/enums/ClaimConditionType.ts
```diff
@@ -0,0 +1,27 @@
+enum ClaimConditionType {
+  StartTime,
+  LimitPerWallet,
+  MerkleRoot,
+  WaitTimeBetweenClaims,
+}
+
+/**
+ * Questions:
+ *
```
jakeloo, line 10: high level, contract specific details:
mint conditions = set the conditions on who can mint the nft.
each condition can be thought of each mint stage with different properties.

there's no limit on how many conditions can be set on the contract. 
the list of conditions has to be sorted by `startTimestamp` in an ascending order and must be unique. 
for example, a list of 3 condtions: [startTimestamp = 0, startTimestamp=30, startTimestamp=60].

src/enums/ClaimConditionType.ts
```diff
@@ -0,0 +1,27 @@
...
+ * 1. Can there only be a single `startTimestamp` condition? What is a `startTimestamp` condition?
+ *
+ * 2. What is `maxMintSupply`?
+ *
+ * 3. What is `currentMintSupply`?
```
jakeloo, line 15: always 0, client should not be setting this. internal to the contract.

src/enums/ClaimConditionType.ts
jakeloo: must be more than 0.

src/enums/ClaimConditionType.ts
jakeloo: 0 = start now

src/enums/ClaimConditionType.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * 4. What is `quantityLimitPerTransaction`?
```
jakeloo, line 17: limit how many NFTs you can mint PER TRANSACTION. 

if it's set to 1, you can only call `claim(1, proofs)` once at a time.

src/enums/ClaimConditionType.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * 5. What is `waitTimeSecondsLimitPerTransaction`?
```
jakeloo, line 19: how long do you have to wait before you can `claim` again.

so if you set `waitTimeSecondsLimitPerTransaction` to MAX_UINT256
you can only `claim` ONCE because you're probably not going to be alive to claim the next one.

if you set it to 15 seconds, you can `claim` every 15 seconds.

if you combine this to MAX_UINT and `quantityLimitPerTransaction` to 1. you'll only be able to claim 1 maximum. in your lifetime at least.

src/enums/ClaimConditionType.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * 6. Can there only be one `pricePerToken`?
```
jakeloo, line 21: yup, pricePerToken = price.

jakeloo: currency = address(0) is the native token
currency = address erc20

src/enums/ClaimConditionType.ts
atbe: > maximum number of mintable nft per condition

What does "per condition" mean in this context?

src/enums/ClaimConditionType.ts
atbe: we should include this as a comment in the contract

src/enums/ClaimConditionType.ts
atbe: > you can only claim ONCE because you're probably not going to be alive to claim the next one.

💀

This makes sense. Composing these conditions is going to be very fun 🤩

src/enums/ClaimConditionType.ts
jakeloo: mint conditions. the array of mint conditions. each individual of the conditions can have different max mint supply

src/enums/ClaimConditionType.ts
jakeloo: yeah, i think we basically want to simplify that in the sense that we'd have a method that does:
max claim quantity per wallet or etc. abstract that away and make it presentable for dev / user.

PR #45, opened 2021-11-22
Renaming

ayshptk:
[diff between `collection.ts` and `bundle.ts`](https://www.diffchecker.com/xBTyOPDK)


src/common/module-type.ts
```diff
@@ -6,7 +6,8 @@ import { $enum } from "ts-enum-util";
  */
 export enum ModuleType {
   CURRENCY = 0,
-  COLLECTION = 1,
+  COLLECTON = 1,
```
nachoiacovino: typo here

src/modules/app.ts
```diff
@@ -233,7 +245,8 @@ export class AppModule extends Module<ProtocolControl> {
   ): Promise<ModuleMetadata[]> {
     const moduleTypesToGet = filterByModuleType || [
       ModuleType.NFT,
-      ModuleType.COLLECTION,
+      ModuleType.COLLECTON,
```
nachoiacovino: same here

nachoiacovino (commented):
Are we also doing the renaming for

NFT -> NFTCollection
Currency -> Token

the same way?

Also, we briefly mentioned with @jakeloo to do the Market -> Marketplace renaming too, is that still happening?

src/common/module-type.ts
ayshptk: this is an outdated  commit, fixed already [here](https://github.com/nftlabs/nftlabs-sdk-ts/blob/1cff35348c3e177b65976c6f39ccb060632e9f9b/src/common/module-type.ts#L9)

src/modules/app.ts
ayshptk: this is an outdated commit, removed it already [here](https://github.com/nftlabs/nftlabs-sdk-ts/blob/1cff35348c3e177b65976c6f39ccb060632e9f9b/src/modules/app.ts#L240)

ayshptk:
@nachoiacovino yep if this is the right way, ill quickly rename the other ones as well. as for the other comments, they are outdated commits do no need to worry about it

src/core/index.ts
```diff
@@ -337,11 +337,14 @@ export class ThirdwebSDK {
 
   /**
    *
-   * @param address - The contract address of the given Collection module.
-   * @returns The Collection Module.
+   * @param address - The contract address of the given Bundle module.
+   * @returns The Bundle Module.
```
jakeloo, line 341: We should properly document deprecation so it shows up in the doc https://tsdoc.org/pages/tags/deprecated/

src/core/index.ts
```diff
@@ -337,11 +337,14 @@ export class ThirdwebSDK {
...
    */
-  public getCollectionModule(address: string): CollectionModule {
-    return this.getOrCreateModule(address, CollectionModule);
+  public getCollectionModule(address: string): BundleModule {
```
jakeloo, line 343: I think we should type alias this. Otherwise, it'd breaking changes for typescripts.

src/modules/bundle.ts
```diff
@@ -14,6 +15,14 @@ import { MetadataURIOrObject } from "../core/types";
 /**
  * @beta
  */
+
+
+export interface BundleMetadata{
```
jakeloo, line 20: we should probably `yarn format` this

src/modules/bundle.ts
```diff
@@ -122,10 +145,22 @@ export class CollectionModule extends ModuleWithRoles<NFTCollectionContract> {
     );
   }
 
-  public async isApproved(address: string, operator: string): Promise<boolean> {
+  public async isApproved(address: string, operator: string, tokenContract?: string, tokenId?: BigNumberish): Promise<boolean> {
```
jakeloo, line 148: I'm confused, what's this `tokenContract` about? 
This is public interfaces, if we want to do it for our internal convenience we should move them to private functions.

jakeloo: @ayshptk we should change this for sure.

PR #49, opened 2021-11-24
fixed decoding error

ayshptk:


src/common/ipfs.ts
```diff
@@ -47,9 +47,12 @@ export async function uploadToIPFS(
...
     headers,
   });
-
-  const body = await res.json();
-  return body.IpfsUri;
+  try {
+    const body = await res.json();
+    return body.IpfsUri;
+  } catch (e) {
+    throw new Error(`Failed to upload to IPFS: ${e}`)
```
atbe, line 54: Throw the `UploadError` here rather than a generic error.

test/nft.test.ts
```diff
@@ -38,10 +44,28 @@ describe("NFT Module", async () => {
...
+    /**
+     * The token with id 1 has been burned and can never be recovered,
+     * so it serves as a good test case.
+     */
+    try {
+      const testBPS = Math.floor(Math.random() * 10) * 100;
+      await nftModule.setRoyaltyBps(testBPS);
+      //throw new Error("should not reach here");
+      //throw new Error(JSON.stringify(await nftModule.getMetadata()));
+      chai.assert.equal(testBPS, (await nftModule.getMetadata()).metadata.seller_fee_basis_points);
```
atbe, line 65: We need to make sure that we're checking that the royalty was updated in *both* places that `setRoyaltyBps` is expected to update. I'll just push this change now because I don't want to hold this PR for any longer, but in the future that's how you would write a thorough test.

src/modules/nft.ts
```diff
@@ -193,9 +193,29 @@ export class NFTModule extends ModuleWithRoles<NFT> {
   }
 
   public async setRoyaltyBps(amount: number): Promise<TransactionReceipt> {
```
atbe, line 195: You got the nft module, nice. But we also need to update the pack, drop, and bundle modules. I'll take care of this, but please note that you only implemented the fix for 1 out of our 4 modules.

Issue #53, opened 2021-11-24
Support permit(...) for gasless approval on tokens


jakeloo:
ERC20 standard doesn't support gasless transactions.
To support gasless transaction, a contract needs to implement EIP-2771, but most ERC-20 contracts out there do not implement it.

There's some ERC-20 tokens out there that supports EIP-2612, for example USDC. EIP-2612 allow users to gasless-ly permit other operators to operate on their funds. Similar behavior from `allowance`.

PR #54, opened 2021-11-24
gasless erc20 permit

jakeloo:
#53 on licketysplit.

requires: https://github.com/nftlabs/ozdefender-autotask/pull/1. will need to communicate that to the customer.


jakeloo:
https://gist.github.com/jakeloo/7b36c341e6e96af4c487ac50d61d1f3b tests

PR #70, opened 2021-12-03
Added New field Support

ayshptk:


src/modules/app.ts
```diff
@@ -339,18 +339,21 @@ export class AppModule
...
-      return Promise.resolve(metadata);
-    }
-    if (metadata.image === undefined) {
-      return Promise.resolve(metadata);
+    for(const key in metadata) {
+      if (typeof metadata[key] === "string") { 
+        return Promise.resolve(metadata);
+      }
+      if (metadata[key] === undefined) { 
+        return Promise.resolve(metadata);
```
atbe, line 347: You should be calling `continue` here, not breaking out of the entire function.

src/modules/app.ts
```diff
@@ -339,18 +339,21 @@ export class AppModule
...
+      }
+  
+      metadata[key] = await uploadToIPFS(
```
atbe, line 350: This is not explicit enough. You need to make sure its a file before uploading it.

src/modules/app.ts
```diff
@@ -339,18 +339,21 @@ export class AppModule
    * @returns - The sanitized metadata with an uploaded image ipfs hash
    */
   private async _prepareMetadata(metadata: CommonModuleMetadata): Promise<any> {
-    if (typeof metadata.image === "string") {
-      return Promise.resolve(metadata);
-    }
-    if (metadata.image === undefined) {
-      return Promise.resolve(metadata);
+    for(const key in metadata) {
+      if (typeof metadata[key] === "string") {
```
atbe, line 343: We need an `if` block dedicated to checking if a property is  a `Dict` or `object`, in which case we need to recursively traverse that object to check its properties, and on and on.

src/modules/app.ts
```diff
@@ -339,18 +339,21 @@ export class AppModule
    * @returns - The sanitized metadata with an uploaded image ipfs hash
    */
   private async _prepareMetadata(metadata: CommonModuleMetadata): Promise<any> {
-    if (typeof metadata.image === "string") {
-      return Promise.resolve(metadata);
-    }
-    if (metadata.image === undefined) {
-      return Promise.resolve(metadata);
+    for(const key in metadata) {
```
atbe, line 342: Break out this logic into its own function so it can be called recursively. Name it `_handleFileProperties` (private)

src/types/module-deployments/CommonModuleMetadata.ts
```diff
@@ -37,4 +37,6 @@ export default class CommonModuleMetadata {
    */
   @JsonProperty("external_link", String, PropertyConvertingMode.IGNORE_NULLABLE)
   externalLink?: string = undefined;
+
+  [key: string]: any;
```
atbe, line 41: What's going on here, are you trying to define the generic `properties` object? Properties aren't really assigned to modules, but we can add it here. Need to make sure we define the property using `json2typescript` like we do the others properties.

src/modules/app.ts
ayshptk: gotcha, but instead of creating a new breakout function why can't we just pass the dict to the `_prepareMetadata`?

src/types/module-deployments/CommonModuleMetadata.ts
ayshptk: yep had to do this to make `metadata[key]` work

src/common/ipfs.ts
```diff
@@ -67,6 +71,21 @@ export async function uploadMetadata(
   if (typeof metadata === "string") {
     return metadata;
   }
+  async function _fileHandler(object: any){
+    const keys = Object.keys(object)
+    for(const key in keys) {
+      if (Buffer.isBuffer(object[keys[key]])) {
+        object[keys[key]] = await uploadToIPFS(object[keys[key]], contractAddress, signerAddress);
+      }
+      if (typeof object[keys[key]] === "object") {
```
jakeloo, line 80: is it possible to do `|| typeof File`? @atbe

atbe: We have a type we can use here called `FileOrBuffer` that should be preferred.

src/common/ipfs.ts
```diff
@@ -67,6 +67,29 @@ export async function uploadMetadata(
...
   }
+  async function _fileHandler(object: any) {
+    const keys = Object.keys(object);
+    for (const key in keys) {
+      if (
+        Buffer.isBuffer(object[keys[key]]) ||
+        (typeof object[keys[key]].name === "string" &&
+          typeof object[keys[key]] !== "object")
+      ) {
+        object[keys[key]] = await uploadToIPFS(
```
atbe, line 78: why is this `[keys[key]]`?

test/ipfs.test.ts
```diff
@@ -0,0 +1,52 @@
...
+        name: "test",
+        image: readFileSync("test/3510820011_4f558b6dea_b.jpg"),
+        test: {
+          test: {
+            image: readFileSync("test/3510820011_4f558b6dea_b.jpg")
+          }
+        }
+      })
+
+      chai.assert.isTrue(upload === "ipfs://bafkreidb4a4h3xg2ju6y3bmkp27ruottvf4gfmjccbtyj5hfk3uypftidu");
```
atbe, line 37: For this test, rather than check the resulting IPFS hash thats returned, I want you to:

1. Upload (you already do this)
2. Download the uploaded data
3. Deserialize it and verify that the `image` and `test.test.image` properties both resolve to images in IPFS

src/common/ipfs.ts
```diff
@@ -67,6 +67,29 @@ export async function uploadMetadata(
   if (typeof metadata === "string") {
     return metadata;
   }
+  async function _fileHandler(object: any) {
+    const keys = Object.keys(object);
+    for (const key in keys) {
+      if (
+        Buffer.isBuffer(object[keys[key]]) ||
```
atbe, line 74: This condition is really obscure. We just need to verify if the property is a `File` or `Buffer`.

I'm pushing a change for this.

atbe:
Also I want to see the following tests:

1. Uploading a `File` object (I think we can only do this in the browser, but I'm not sure). If it can only be tested in the browser, please verify the behavior within console. Also make sure the actions you take in console *actually test the code paths we've modified*.
2. Uploading a `Blob` object
3. Uploading recursive `File || Blob` properties. You already have this in the tests, but I want to see better verification instead of just checking the hash.

ayshptk:
browser uploading works: tokenId is 4 [in this contract](https://thirdweb-www-git-test-uploads-thirdweb.vercel.app/mumbai/0xeef5Db17cA4e5333D99d47d2c6297E01949A3491/nft-collection/0x79bE3C3542c189245963c93A2600eAb106E99F1A)

using [`v1.19.3-0`](https://github.com/nftlabs/nftlabs-sdk-ts/releases/tag/v1.19.3-0)

PR #77, opened 2021-12-08
Claim condition

jakeloo:


atbe:
- [ ] Batch fetching of tokens in a drop
- [ ] Easier accessing of current phases with easy to understand accessors for current phase
- [ ] `canClaim` method to check if user can claim a drop
- [ ] If a users cannot claim, provide a method that can return a friendly readable reason why
- [ ] Response codes constants file that allows you to translate an error to a friendly message

atbe:
Abandoning in favor of #88

PR #78, opened 2021-12-08
Multiple Gateway Support [NEED FEEDBACK]

ayshptk:


jakeloo (changes requested):
I think we should take this opportunity to clean up the interface and make it more future proof.
1. don't hard code timeout
2. don't hard code ipfs gateway urls

PR #79, opened 2021-12-08
Implementing `setAllRoleMembers` [NEED FEEDBACK]

ayshptk:


src/core/module.ts
```diff
@@ -521,6 +520,50 @@ export class ModuleWithRoles<
     return roles;
   }
 
+  public async setAllRoleMembers(
```
atbe, line 523: Please add a super clear docstring that describes how this method works.

We should especially call out that this method is a **replace** for all roles currently assigned to a module and that any roles that you *dont want removed* should always be included, otherwise folks will accidentally remove themselves and the module will be lost forever.

ayshptk: > that any roles that you dont want removed should always be included

not really, the function only replaces the roles present in the object they passed. if they dont pass a role, the role will not be changed. i think that has more utility than the other way because most use cases would want to overwrite just one role.

atbe: Sure makes sense. This PR LGTM, backmerge `main` and resolve the conflicts and assuming all tests pass then I'll merge it in.

PR #94, opened 2021-12-15
Get all

ayshptk:


src/modules/token.ts
```diff
@@ -124,6 +125,32 @@ export class CurrencyModule extends ModuleWithRoles<Coin> {
     }
     await this.sendTransaction("multicall", [encoded]);
   }
+  public async getAll(): Promise<Record<string, BigNumber>> {
```
atbe, line 128: This function needs a docstring. Just from reading it, I can say that the name doesn't communicate what it is doing.

`getAll` doesn't seem like the right name for this method. It looks like its fetching all balances and holders of a token.

ayshptk:
cc @atbe

test/token.test.ts
```diff
@@ -57,4 +58,10 @@ describe("Token Module", async () => {
       );
     }
   });
+  it("Should parse events properly", async () => {
```
atbe, line 61: This test is not super clear. The name is useless too.

The test title should be `should return the correct list of holders and their balances`

The test should:

1. Transfer tokens to multiple wallets (at least 5 wallets)
2. Call the `getAllHolderBalances` method and check the balance of every single wallet that we expect the tokens to be sent to.

PR #97, opened 2021-12-15
new specific errors, exposed `setRestrictedListerRoleOnly` and `token…

ayshptk:
…sPerBuy`


src/common/error.ts
```diff
@@ -29,6 +29,35 @@ export class UploadError extends Error {
...
+}
+
+export class NotOwner extends Error {
+  /** @internal */
+  constructor() {
+    super(`LIST ERROR: you should be the owner of the token to list it.`);
+  }
+}
+
+export class NotAuthorised extends Error {
```
atbe, line 46: Spell it `NotAuthorized`

src/common/error.ts
```diff
@@ -29,6 +29,35 @@ export class UploadError extends Error {
...
+  /** @internal */
+  constructor() {
+    super(`LIST ERROR: you must list at least one token.`);
+  }
+}
+
+export class NotOwner extends Error {
+  /** @internal */
+  constructor() {
+    super(`LIST ERROR: you should be the owner of the token to list it.`);
```
atbe, line 42: these errors will be used throughout the sdk, so don't hardcode the message. Instead, have a default value, and allow a custom message to be passed in the constructor. Look at the other errors for an example.

src/common/error.ts
```diff
@@ -29,6 +29,35 @@ export class UploadError extends Error {
...
+  constructor() {
+    super(
+      `LIST ERROR: you do not have the permission to list in the marketplace`,
+    );
+  }
+}
+export class BuyLimit extends Error {
+  /** @internal */
+  constructor(message: string) {
+    super(`BUY ERROR: You cannot buy more tgan ${message} tokens`);
```
atbe, line 57: `tgan` is a typo

atbe: and in this case, `message` should not be the quantity, otherwise you'd name it `quantity`.

Here's the errors I'm thinking about:


- `MissingRoleError` -> Throw this when someone should have a role, but is missing that role. It should accept 2 parameters:
    - `address` the address thats missing the permission
    - `role` the role that the address is missing
- `AssetNotFoundError` -> The asset you're trying to use could not be found. If we're passing in an asset contract address in the market list method, we should check to see if it exists, if not, throw this error and include the asset address in the error message
- `NotEnoughTokensError` -> Anywhere someone tries to perform an action where they're trying to use more tokens than they have, throw this error. Parameters include:
    - The address of the assets contract
    - The quantity that was requested
    - Optionally, the amount available (which should be less than needed, otherwise this error would not be thrown)

src/common/error.ts
```diff
@@ -29,6 +29,35 @@ export class UploadError extends Error {
   }
 }
 
+export class NoTokenListed extends Error {
```
atbe, line 32: This shouldn't be its own error class. Make this an `invariant` call in the `list` method instead.

atbe:
@ayshptk please look at the `lgtm-com` bot comment, you're missing some `await` calls.

ayshptk:
mb @atbe

atbe:
@ayshptk assuming you fixed those names for those error classes, please get this merged.

PR #109, opened 2021-12-21
resolve recursively

ayshptk:


src/common/contract.ts
```diff
@@ -58,14 +58,13 @@ export async function getContractMetadata(
   const uri = await contract.contractURI();
   const gatewayUrl = replaceIpfsWithGateway(uri, ipfsGatewayUrl);
   const meta = await fetch(gatewayUrl);
+  let json = await meta.json();
```
jakeloo, line 61: why did we move this outside try catch?

PR #122, opened 2021-12-23
Set royalty treasury on deployment 

ayshptk:


src/modules/app.ts
```diff
@@ -443,8 +443,14 @@ export class AppModule
       ],
       NFTCollection__factory,
     );
+    if (
+      metadata.feeRecipient !== undefined &&
+      metadata.feeRecipient !== address
```
jakeloo, line 448: this should be protocol control / app address not the deploying module's address.

jakeloo (changes requested):

jakeloo (approved):

PR #126, opened 2021-12-24
Restricted Transfer

ayshptk:


src/modules/drop.ts
```diff
@@ -1406,12 +1408,6 @@ class DropV1Module extends ModuleWithRoles<Drop> {
     return await this.sendTransaction("setMaxTotalSupply", [amount]);
   }
 
-  public async setRestrictedTransfer(
```
jakeloo, line 1409: why is this nuked?

jakeloo (changes requested):
also, docstring would be nice.

src/modules/pack.ts
```diff
@@ -347,7 +347,9 @@ export class PackModule extends ModuleWithRoles<PackContract> {
   public async setRestrictedTransfer(restricted = false): Promise<void> {
     await this.sendTransaction("setRestrictedTransfer", [restricted]);
   }
-
+  public async isRestrictedTransfer(): Promise<boolean> {
```
atbe, line 350: We should rename all instances of `isRestrictedTransfer` to `isTransferRestricted`.

src/modules/drop.ts
atbe: moved to common interface, method name is still the same.

PR #128, opened 2021-12-25
updated docs

rodae:
minor typo fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@3rdweb/sdk",
-  "version": "1.29.4",
+  "version": "1.29.5",
   "description": "The main thirdweb SDK.",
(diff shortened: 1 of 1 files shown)


atbe:
hi @rodae thanks so much for the contribution! Our docs are actually autogenerated, so they're not something we edit directly (otherwise the CI pipeline will overwrite your changes, oops ✍️ )

For us to merge this, could you make the change here instead? https://github.com/rodae/nftlabs-sdk-ts/blob/patch-1/src/index.ts#L29 That's the line where this docstring is generated from.

Thx so much, and wishing you happy holidays!

rodae:
Hey @atbe thanks for the heads up. Is it ok to add the commit to this pull request or should I close and open a new one?

PR #133, opened 2021-12-26
Flag for bulk uploads

ayshptk:


src/storage/IpfsStorage.ts
```diff
@@ -63,8 +65,23 @@ export class IpfsStorage implements IStorage {
...
+    contractAddress?: string,
+  ): Promise<string> {
+    const cid = await this.uploadBatchWithCid(true, files, contractAddress);
+
+    return `ipfs://${cid}/`;
+  }
+
   private async uploadBatchWithCid(
-    files: Buffer[] | string[] | FileOrBuffer[] | File[],
+    withFileNames = false,
```
atbe, line 78: Please put this argument at the end of the function parameter list so that it doesn't disrupt existing function calls.

This is a common programmer mistake when trying to retain backwards compatibility. Here's an example.

Say there's a function:

```ts
function add(x: number, y: number, optionA = false)
```

Lets say a developer *John* used this code and called this function like this:

```ts
add(1,2, true)
```

If we add a parameter to the front of the function like this, calls to the function will still work **but every single instance where the function was called needs to be updated because the ordering has changed**.

```js
function add(x: number, y: number, newOption = false, optionA = false)
```

Lets look back at that Johns code:

```js
add(1,2, true)
```

Now John doesn't realize it, but he's accidentally passing `true` for `newOption` because it took the position of `optionA` in the arg list.

---

The backwards compatible change would be to **preserve the ordering of existing parameters** like this:


```js
function add(x: number, y: number, optionA = false, newOption = false)
```

src/storage/IpfsStorage.ts
```diff
@@ -63,8 +65,23 @@ export class IpfsStorage implements IStorage {
...
+    files:
```
atbe, line 79: I don't like this approach with the `files` field taking on so many different types in the union.

Move on to another ticket for now and I can come in and clean this interface up.

src/types/FireOrBufferWithNames.ts
```diff
@@ -0,0 +1,6 @@
+type FileOrBufferWithNames = {
+  file: Buffer | File;
+  name: string;
+};
+
+export default FileOrBufferWithNames;
```
atbe, line 6: Don't do any `default` exports, it causes issues when exporting the type at the module level. Instead just an `export` is fine.

src/storage/IpfsStorage.ts
ayshptk: Gotcha thanks! But I thought about this and it is a private function and we need `fileStartNumber` when `withFileNames` is true. so `withFileNames` has to be before `fileStartNumber` right?

test/ipfs.test.ts
```diff
@@ -108,6 +109,22 @@ describe("IPFS Uploads", async () => {
...
+      { file: readFileSync("test/test.mp4"), name: "test3.mp4" },
+      {
+        file: readFileSync("test/test.mp4"),
+        name: "test1.mp4",
+      },
+    ];
+    const cid = await sdk.getStorage().uploadBatchWithFileNames(sampleObjects);
+    console.log(cid);
+  });
+
```
atbe, line 127: Two things we need in this test suite in order to merge:

1. [ ] A test for the `uploadBatch` that:
    - [ ] Calls upload with `uploadFileStartNumber` = to something > 0, then ensures that fetching all files at valid indexes works (like `{cid}/10`, `{cid}/11` ...)
    - [ ] Calls upload with `upoadFileStartNumber` = to 0, and checks all fetched files
2. [ ] Update the `should upload files with filenames correctly` test to actually fetch the files using their file names and ensures correct content
    - [ ] Fetching a `{cid}/test1.mp4` should work

src/interfaces/IStorage.ts
```diff
@@ -91,4 +92,9 @@ export interface IStorage {
     contractAddress?: string,
     fileStartNumber?: number,
   ): Promise<string>;
+
+  uploadBatchWithFileNames(
```
atbe, line 96: Docstring please!

src/types/FireOrBufferWithNames.ts
```diff
@@ -0,0 +1,4 @@
+export type FileOrBufferWithNames = {
+  file: Buffer | File;
+  name: string;
```
atbe, line 3: Lets make this field optional.

With this change, when uploading the objects please make these changes:

- [ ] If the object is a `File` object && the `name` field is `undefined` use the `File.name` property
- [ ] If the object is a `File` object && `name` field is not `undefined` use the passed file name
- [ ] If the object is not a `File` object && the `name` field is undefined, throw an error (`FileNameMissingError`)

jakeloo:
1. i don't think we should have strict file.name check i.e: `=== undefined` 
because it someone passes in `""` as file name is that really valid?

2. should file name be unique? can i pass in `[{name: "a"}, {name: "a"}, {name: "A"}]`?

Issue #117, opened 2021-12-22
Create explicit return type when calling a tx that writes data which includes the resulting tx hash1


atbe:
Customers to ping when done:
- https://discord.com/channels/834227967404146718/923160861705642004/923290228452655136

PR #152, opened 2021-12-31
Return transactions

embiem:
I'd like to propose to return the transactions, for use cases like when we want to show the transaction hash to the user.

jake agreed to do it here: https://discord.com/channels/834227967404146718/923160861705642004/923290228452655136, so I figured creating a PR would be fine.

Cheers and happy new year!


atbe:
@embiem thanks for the contribution! How do you feel about returning an interface like this instead?

```js
interface TransactionWithResult<T = void> {
  result?: T;
  transactionHash: string;
}
```

Reason being that we want to start returning transaction hashes *everywhere* in the SDK, and this interface can provide a good base for us to return common things like the tx hash (and at some point, the estimated gas cost, etc).

---

The `bundle.ts` `mind` method, for example (assuming `mintTo` returns :

```js
  public async mintTo(
    to: string,
    args: INFTBundleBatchArgs,
    data: BytesLike = [0],
  ): Promise<TransactionWithResult> {
    return {
      transactionHash: await this.sendTransaction("mint", [to, args.tokenId, args.amount, data]),
    };
  }
```

This can lay a good foundation for us and avoids the lock-in of returning *only* the transaction has in some methods, and *only* returning some data in other methods.

atbe:
Or taking it one step further, two explicit types:

```js
export interface Transaction {
  transactionHash: string;
  // In here is where we'll include future info about tx's thats common to them all
}

export interface TransactionWithResult<T> extends Transaction {
  result?: T;
}
```

So that its **extra** clear when a method is returning some result.

atbe:
related #117

embiem:
That'd be even better. I just wanted to add the return, but your suggestion would be the proper approach.

embiem:
@atbe is it OK if I add it to this PR? Or do you want to include it in a different one?

jnsdls:
hey @embiem, sorry for the lag on getting this in, we're actively working on a `v2` of the sdk which will include this style of transaction interface for every "write" action

_something close to_
```typescript
type ReturnTypeFromTransaction = {
  transaction: ethers.TransactionReceipt,
  result: Promise<resultDataType>
}
```
we're modeling it off of the `fetch` return interface

```typescript
const res = await fetch(...);
const data = await res.json() // for example
```

so in v2 a `mint` would look something like this for example:
```typescript
const mintRes = await module.mint({...});
// if you just need the transaction hash
const txnHash = mintRes.transaction.hash;
// or if you need the data
const resultData = await mintRes.result();
```

you can check out the ongoing work on this on the `v2` branch if you're interested, we're hoping to get something ready for release in the next coming days

embiem:
Oh awesome! Then we can close this one.

PR #153, opened 2021-12-31
Set Contract URI only when it is changed

ayshptk:


src/common/isMetadataEqual.ts
```diff
@@ -0,0 +1,15 @@
+export function isMetadataEqual(old: any, changed: any): boolean {
```
jakeloo, line 1: we should @internal this, so it doesn't show up in docs.

src/modules/drop.ts
```diff
@@ -490,26 +497,28 @@ export class DropModule
     factory.allSnapshots().forEach((s) => {
       merkleInfo[s.merkleRoot] = s.snapshotUri;
     });
+    const encoded = [];
     const { metadata } = await this.getMetadata(false);
     invariant(metadata, "Metadata is not set, this should never happen");
     if (factory.allSnapshots().length === 0 && "merkle" in metadata) {
       metadata["merkle"] = {};
-    } else {
+    } else if (!isMetadataEqual(merkleInfo, metadata["merkle"])) {
```
jakeloo, line 505: i think this is logic bug.

we should not do `else if` because if the code wants to clear out merkle, then it should be able to.
we need to handle the case (as part of testing)
no merkle -> has merkle -> same merkle -> updates merkle -> no merkle

ayshptk: what if we just see if there's a change in `metadata["merkle"]` by comparing it with a copy just before processing? it's just strings so I dont think that would be an issue  - [commit](https://github.com/nftlabs/nftlabs-sdk-ts/pull/153/commits/37e4116b9a86f8ede1fdef1515e4b612a55017f2)

src/modules/drop.ts
```diff
@@ -490,26 +497,33 @@ export class DropModule
     factory.allSnapshots().forEach((s) => {
       merkleInfo[s.merkleRoot] = s.snapshotUri;
     });
+    const encoded = [];
     const { metadata } = await this.getMetadata(false);
     invariant(metadata, "Metadata is not set, this should never happen");
+    const oldMerkle = JSON.parse(JSON.stringify(metadata["merkle"]));
```
jakeloo, line 503: why do we need to stringify then parse it? seems unnecessary?

jakeloo (changes requested):

src/modules/drop.ts
jakeloo: yeah that's the idea, compare metadata["merkle"] to see if they're the same or different

jakeloo (approved):

PR #154, opened 2021-12-31
Withdrawal Bug Fixes

ayshptk:


src/modules/app.ts
```diff
@@ -344,11 +344,33 @@ export class AppModule
   public async withdrawFunds(
     to: string,
     currency: string,
-  ): Promise<TransactionReceipt> {
+  ): Promise<TransactionReceipt | void> {
```
ayshptk, line 347: @jakeloo will this be an issue?

src/modules/app.ts
```diff
@@ -901,6 +923,29 @@ export class AppModule
...
+  }
+
+  public async upgrade(): Promise<void> {
+    if (!(await this.requireUpgrade())) {
+      return;
+    }
+    const metadata = (await this.getMetadata()).metadata;
+    const splitModule = await this.deploySplitsModule({
+      name: `${metadata?.name} Royalty Treasury`,
+      recipientSplits: [],
```
jakeloo, line 945: we want to add signer in here

src/modules/app.ts
jakeloo: yes, let's make it TransactionReceipt

src/modules/app.ts
```diff
@@ -910,7 +955,10 @@ export class AppModule
     const walletBalance = await this.readOnlyContract.provider.getBalance(
       this.address,
     );
-    return walletBalance;
+    const treasuryBalance = await this.readOnlyContract.provider.getBalance(
+      await this.getRoyaltyTreasury(),
+    );
+    return walletBalance.add(treasuryBalance);
```
jakeloo: need similar logic in `balanceOfToken`

src/modules/app.ts
```diff
@@ -344,11 +344,33 @@ export class AppModule
...
+    const treasury = await this.getRoyaltyTreasury();
+    if (
+      !(await erc20.balanceOf(treasury)).isZero() &&
+      treasury !== this.address
+    ) {
+      const royalty = Royalty__factory.connect(
+        treasury,
+        this.readOnlyContract.provider,
+      );
+      return await this.sendContractTransaction(
```
jakeloo, line 367: double check the logic in splits module, i think depending on the currency token address, we need to do call differently

jakeloo (changes requested):

PR #176, opened 2022-01-06
`useSnapshot` function [WIP]

ayshptk:


src/factories/ClaimConditionPhase.ts
```diff
@@ -100,14 +102,24 @@ export default class ClaimConditionPhase {
     return this;
   }
 
+  /**
```
jakeloo, line 105: we don't need to deprecate it because `await` on non-promise is not breaking.

src/factories/ClaimConditionFactory.ts
```diff
@@ -141,16 +148,23 @@ class ClaimConditionFactory {
...
+    const sorted = await this.buildConditions();
     const cleared = sorted.splice(index - 1, 1);
     this.fromPublicClaimConditions(cleared);
   }
 
+  /**
+   * @deprecated - Use {@link deleteClaimPhase} instead.
+   */
+  public removeClaimPhase(_index: number): void {
+    throw new FunctionDeprecatedError("deleteClaimPhase");
```
jakeloo, line 165: not too big of a fan of this. in general we need to keep deprecated function working.
one thought is that we can offload the removed index until the end, build conditions time.

src/factories/ClaimConditionFactory.ts
```diff
@@ -47,32 +49,37 @@ class ClaimConditionFactory {
...
-        const now = Math.floor(Date.now() / 1000);
-        return {
-          ...c,
-          startTimestamp: c.startTimestamp.lt(now)
-            ? BigNumber.from(0)
-            : c.startTimestamp.sub(now),
-        };
+    let sorted: PublicClaimCondition[] = [];
+    await Promise.all(
+      this.phases.map(async (c) => await c.buildPublicClaimCondition()),
```
jakeloo, line 57: you don't need `async (c) => await` here because `Promise.all` will resolve the promises in the array

jakeloo (changes requested):

src/factories/ClaimConditionFactory.ts
ayshptk: could you elaborate a bit more?

PR #191, opened 2022-01-11
add initial generate json snippets fn

jnsdls:


jnsdls:
@atbe @jakeloo this basically adds a bunch of code examples & other descriptions & documentation that then gets built into a `docs/snippets.json` with `yarn build`

that `docs/snippets.json` -> gets pulled in by `console` at run-time to generate the `Code` section on modules (& we may also re-use this for the code snippets page in portal)

PR #192, opened 2022-01-11
Set balance in ownedByAddress for each token returned in getOwned

mcintyre94:
This PR updates the bundle/collection module `getOwned` function so that it returns the number of each token owned by the user in the `ownedByAddress` field. Currently it returns 0 in this field for all tokens. This allows consumers to get the metadata and balance of every owned token in a bundle for a given address in a single call to `getOwned`. 

This was discussed on Discord and agreed as a bug (partner channel link): https://discord.com/channels/834227967404146718/915738548848713818/929025948999884901

Note that because `ownedByAddress` holds a number (not BigNumber), I've added an additional check and we return `Number.MAX_SAFE_INTEGER` if the number owned is greater than this number.

### DIFF
--- a/src/modules/bundle.ts
+++ b/src/modules/bundle.ts
@@ -22,7 +22,7 @@ export interface BundleMetadata {
   creator: string;
   supply: BigNumber;
   metadata: NFTMetadata;
-  ownedByAddress: number;
+  ownedByAddress: BigNumber;
   underlyingType: UnderlyingType;
 }
 
@@ -118,7 +118,7 @@ export class BundleModule
       creator: state.creator,
       supply,
       metadata,
-      ownedByAddress,
+      ownedByAddress: BigNumber.from(ownedByAddress),
       underlyingType: state.underlyingType,
     };
   }
@@ -459,7 +459,10 @@ export class BundleModule
       })
       .filter((b) => b.balance.gt(0));
     return await Promise.all(
-      ownedBalances.map(async (b) => await this.get(b.tokenId.toString())),
+      ownedBalances.map(async ({ tokenId, balance }) => {
+        const token = await this.get(tokenId.toString());
+        return { ...token, ownedByAddress: balance };
+      }),
     );
   }
 
--- a/test/collection.test.ts
+++ b/test/collection.test.ts
@@ -59,6 +59,24 @@ describe("Bundle Module (aka Collection Module)", async () => {
       .length(0, "Bob should not have any nfts");
   });
 
+  it("should return non-zero balance of owned collection tokens", async () => {
+    const token = await nftModule.mint({
+      name: "test",
+    });
+
+    try {
+      await collectionModule.createWithERC721(nftModule.address, token.id, {
+        name: "TEST NFT",
+      });
+    } catch (err) {
+      assert.fail(err);
+    }
+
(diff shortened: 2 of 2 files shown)


src/modules/bundle.ts
```diff
@@ -459,7 +459,14 @@ export class BundleModule
       })
       .filter((b) => b.balance.gt(0));
     return await Promise.all(
-      ownedBalances.map(async (b) => await this.get(b.tokenId.toString())),
+      ownedBalances.map(async ({ tokenId, balance }) => {
+        const token = await this.get(tokenId.toString());
+        if (balance.gt(Number.MAX_SAFE_INTEGER.toString())) {
+          return { ...token, ownedByAddress: Number.MAX_SAFE_INTEGER };
```
jnsdls, line 465: should never work with straight `number` types. `balance` is a `BigNumber` type and *should* be a `BigNumber` type for that reason. can be rendered as `toString()` in the client and can use `BigNumber` math functions to do math with

src/modules/bundle.ts
```diff
@@ -459,7 +459,14 @@ export class BundleModule
...
+        } else {
+          return { ...token, ownedByAddress: balance.toNumber() };
```
jnsdls, line 467: changing the above means that this if / else statement becomes unnecessary

jnsdls (changes requested):
small change to use the `BigNumber` type of the `balance` directly instead of converting it to be an integer

src/modules/bundle.ts
mcintyre94: I'm not sure exactly how I can do that though, because `ownedByAddress` is specified as a `number` instead of `BigNumber` - so don't I have to convert to a `number` to store on that field?

```
export interface BundleMetadata {
  creator: string;
  supply: BigNumber;
  metadata: NFTMetadata;
  ownedByAddress: number;
  underlyingType: UnderlyingType;
}
```

I do agree that it'd be better as a `BigNumber` but I'm not sure how else I can store it on `ownedByAddress` given the existing type there.

mcintyre94: I've pushed a change that makes the type of `ownedByAddress` be `BigNumber` which makes everything just work. I've made it a separate commit because I haven't run all the tests locally yet (collection.test.ts all pass) and I'm not sure whether you'll want that type being changed.

test/collection.test.ts
```diff
@@ -59,6 +59,24 @@ describe("Bundle Module (aka Collection Module)", async () => {
...
+      await collectionModule.createWithERC721(nftModule.address, token.id, {
+        name: "TEST NFT",
+      });
+    } catch (err) {
+      assert.fail(err);
+    }
+
+    const nfts = await bundleModule.getOwned(adminWallet.address);
+    expect(nfts).to.be.an("array").length(1);
+    assert.equal(nfts[0].ownedByAddress, 1);
```
jakeloo, line 77: you might have to do `equal(nfts[0].ownedByAddress.eq(1), true)` for equality between bignumber and a number

mcintyre94: Thanks Jake, updated this to `assert.isTrue(nfts[0].ownedByAddress.eq(1));`

src/modules/bundle.ts
jnsdls: this LGTM

PR #201, opened 2022-01-13
fixed bug which did not respect ipfs urls

ayshptk:


src/storage/IpfsStorage.ts
```diff
@@ -80,7 +80,8 @@ export class IpfsStorage implements IStorage {
       | string[]
       | FileOrBuffer[]
       | File[]
-      | BufferOrStringWithName[],
+      | BufferOrStringWithName[]
+      | string[],
```
jakeloo, line 84: remove string[]

src/storage/IpfsStorage.ts
```diff
@@ -99,7 +100,8 @@ export class IpfsStorage implements IStorage {
...
```
jakeloo, line 104: remove string[]

src/storage/IpfsStorage.ts
```diff
@@ -109,7 +111,7 @@ export class IpfsStorage implements IStorage {
     };
     const data = new FormData();
     const fileNames: string[] = [];
-    files.forEach((file, i) => {
+    files.forEach(async (file, i) => {
```
jakeloo, line 114: remove async

src/storage/IpfsStorage.ts
```diff
@@ -313,8 +319,18 @@ export class IpfsStorage implements IStorage {
   ) {
     const finalMetadata: MetadataURIOrObject[] =
       await this.batchUploadProperties(metadatas);
+    const filesToUpload = await Promise.all(
```
jakeloo, line 322: we shouldn't need these and these can be kept the same

jakeloo (changes requested):

PR #203, opened 2022-01-13
resolves ens domains in snapshot

ayshptk:


jakeloo:
we can't do this because it'd hit rate limit

adam-maj:
Is this no longer in review @jakeloo ?

jakeloo:
Going to close this for now, we'll re-open it when we have a better solution for dealing with provider rate limit issue.

PR #204, opened 2022-01-13
added validation

ayshptk:


adam-maj:
@atbe  @ayshptk whats the status on this one

src/core/index.ts
```diff
@@ -514,6 +515,22 @@ export class ThirdwebSDK implements IThirdwebSdk {
   }
 
   public async createSnapshot(leafs: string[]): Promise<SnapshotInfo> {
+    const counts: { [key: string]: any } = {};
```
jakeloo, line 518: we should change this to if there's non valid address, we should throw an error. so it's more similar to current behavior where when there's duplicate there'll be error.

jakeloo (changes requested):

test/drop.test.ts
```diff
@@ -256,6 +257,55 @@ describe("Drop Module", async () => {
...
+      abbyWallet.address,
+      w1.address,
+      w2.address,
+      w3.address,
+      w4.address,
+      w4.address,
+    ];
+
+    const hashedLeafs = members.map((l) => keccak256(l));
+    const tree = new MerkleTree(hashedLeafs, keccak256, {
```
atbe, line 273: Do you actually use this `tree` variable anywhere? If not, lets remove it.

src/core/index.ts
```diff
@@ -514,6 +515,22 @@ export class ThirdwebSDK implements IThirdwebSdk {
...
+    const repeated: string[] = [];
+    leafs.forEach((leaf) => {
+      leaf = leaf.toLowerCase();
+      invariant(isAddress(leaf), "Invalid address : " + leaf);
```
atbe, line 522: Instead of using `invariant` here, lets throw custom error types so they're easier to catch.

src/core/index.ts
```diff
@@ -514,6 +515,22 @@ export class ThirdwebSDK implements IThirdwebSdk {
...
+      if (counts[leaf] && repeated.indexOf(leaf) === -1) {
+        repeated.push(leaf);
+      } else {
+        counts[leaf] = 1;
+      }
+    });
+    invariant(
```
atbe, line 529: Instead of using `invariant` here, lets throw custom error types so they're easier to catch.

src/core/index.ts
ayshptk: hmm already checks for both invalid address and repeated addresses, tho added customer errors instead of `invariants` rn

src/common/error.ts
```diff
@@ -23,6 +23,15 @@ export class InvalidAddressError extends Error {
   }
 }
 
+export class RepeatedAddressError extends Error {
+  /** @internal */
+  constructor(address?: string) {
+    super(
+      address ? `'${address}' is was passed more than once` : "An address was passed more than once",
```
atbe, line 30: `is was`

>>>

PR #207, opened 2022-01-14
Events: Added event to emit when gas-less transaction is signed

ximxim:
As part of a project that I am working on, we want to show a different UI state when the transaction is signed vs when the transaction is pending signature. 

I added EventEmitter to trigger an event that the consumer to listen to.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "@types/chai": "^4.2.22",
     "@types/crypto-js": "^4.0.2",
     "@types/deep-equal-in-any-order": "^1.0.1",
+    "@types/events": "^3.0.0",
     "@types/expect": "^24.3.0",
     "@types/form-data": "^2.5.0",
     "@types/mocha": "^9.0.0",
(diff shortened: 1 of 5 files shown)


atbe:
Looks great, thanks for the contribution @ximxim

Let us know when you're setup the constants file with the event keys. We'll want to make sure they're properly exported so folks outside the sdk can import them.

Looking forward to merging this, there are lots of spots we can emit events in the sdk.

ximxim:
@atbe updated the PR with event names extracted to its file and exposed to SDK consumers.

atbe:
This is great @ximxim 

One last request, if you don't mind. Could you emit an event when the biconomy request succeeds? We can emit the transaction hash that's returned from biconomy. The response text from biconomy looks like this:

```json
{
  "code": 200,
  "message": "Meta transaction sent to blockchain",
  "txHash": "...",
  "flag": 200,
  "log": "Meta transaction sent to blockchain",
  "retryDuration": 158
}
```

Emitting the `txHash` property would be very valuable.

PR #212, opened 2022-01-17
am/open-marketplace

adam-maj:


jakeloo:
Can we make the naming more explicit than `open`? 
`setRestrictedToListerRole` or something?

src/modules/marketplace.ts
```diff
@@ -1124,4 +1124,12 @@ export class MarketplaceModule
       | DirectListing
     )[];
   }
+
+  public async isRestrictedListerRoleOnly(): Promise<boolean> {
+    return this.readOnlyContract.restrictedListerRoleOnly();
+  }
+
+  public async setRestrictedListerRoleOnly(isOpen: boolean): Promise<void> {
```
joaquim-verges, line 1132: this signature seems wrong - `setRestricted...()` but passes `isOpen` which is the opposite of restricted?

should be `setRestrictedToListerRole(isRestricted: boolean)` (which is what the contract expects)

PR #215, opened 2022-01-19
await all of the calls to 'setModuleRoyaltyTreasury'

atbe:


atbe:
@ayshptk for visibility, lets not let bugs like these slide again in the future 🙏

atbe:
thanks @nkrishang

PR #223, opened 2022-01-21
Date resolver

ayshptk:


atbe:
11 marketplace tests are failing, please get those fixed so we can merge this in.

jakeloo:
we should decouple this PR with the storage change from the other PR

test/before.test.ts
```diff
@@ -47,6 +47,11 @@ before(async () => {
     ipfsGatewayUrl,
     registryContractAddress: registryAddress,
     maxGasPriceInGwei: 10000,
+    storage: {
+      thirdwebIpfsServerUrl:
```
atbe, line 51: Why is this stuff in this PR? This PR should be **only** about the date resolver, that's it. Can't merge this with these changes.

PR #225, opened 2022-01-21
Storage Options

ayshptk:


jakeloo:
I don't think we should approach it this way. This feels like a backward design. We already setup `IStorage` with the intention of letting users overwriting the Storage, so why are we leaking the implementation details up to SDK options?

If the issue that we're tackling is that our tests are using up production pinata api rate limit, then let's fix that in the test.
In the test, we can overwrite the Storage class to be either a Mock storage that's solely upload/download in memory OR a copy-paste of existing IpfsStroage but with the URL swapped.

PR #229, opened 2022-01-24
ClaimTo Bug

ayshptk:


src/modules/drop.ts
```diff
@@ -870,20 +870,18 @@ export class DropModule
...
-      ]),
-    );
-    encoded.push(
-      this.contract.interface.encodeFunctionData("transferFrom", [
-        await this.getSignerAddress(),
-        addressToClaim,
-        (await this.readOnlyContract.nextTokenIdToMint()).sub(1),
-      ]),
-    );
+    const result = await this.claim(quantity, claimData.proofs);
```
jnsdls, line 873: this will make it not be a multi-call?

jnsdls (commented):
hm, unsure about this?

src/modules/drop.ts
atbe: Yeah, this is expected. Because we don't know the ID's of the NFTs that are about to be claimed, we need to form the `claim` in one tx, then the transfer calls in a separate (multicall) tx.

@jnsdls

jakeloo:
@atbe wasn't aware that this pr been up. i've made similar changes on my PR but it doesn't go through gateway:  
https://github.com/nftlabs/nftlabs-sdk-ts/pull/231/files#diff-b9323da2530c9cbff5e0bcab36a37c2191d93fd1ca8e99ec3c36cffaef2f8303

PR #231, opened 2022-01-25
contracts upgrades

jakeloo:
edit: wasn't aware that #229 fixed the issue with `claimTo`, but i've made similar fix in this pr.


jakeloo:
DO NOT MERGE yet. will run through a test on dashboard.

PR #232, opened 2022-01-27
Port Drop module read functions

joaquim-verges:


src/modules/drop-erc-721.ts
```diff
@@ -10,6 +14,20 @@ import {
 } from "../schema/modules/drop-erc721";
 import { SDKOptionsOutput } from "../schema/sdk-options";
 import { ContractRoyalty } from "../core/classes/contract-royalty";
+import { BigNumber } from "ethers";
+import { AddressZero } from "@ethersproject/constants";
+import { ClaimCondition } from "../types";
+
+// TODO zod-ify
+type NFTMetadataOwner = {};
+type NFTMetadata = {};
```
joaquim-verges, line 23: @jnsdls needs zodification

src/modules/drop-erc-721.ts
```diff
@@ -10,6 +14,20 @@ import {
...
+
+// TODO extract
+const DEFAULT_QUERY_ALL_COUNT = 100;
+type QueryAllParams = {
+  start: number;
+  count: number;
+};
```
joaquim-verges, line 30: not sure where to put this, we re-use it in other places right?

src/modules/drop-erc-721.ts
```diff
@@ -56,7 +75,305 @@ export class DropErc721Module {
...
+      ),
+    );
+    const unmintedId =
+      await this.contractWrapper.readContract.nextTokenIdToClaim();
+    return await Promise.all(
+      Array.from(Array(maxId.sub(unmintedId).toNumber()).keys()).map((i) =>
+        this.getTokenMetadata(unmintedId.add(i).toString()),
+      ),
+    );
+  }
```
joaquim-verges, line 182: these getAllX() functions could be refactored a bit, but leaving as is for now

src/modules/drop-erc-721.ts
```diff
@@ -56,7 +75,305 @@ export class DropErc721Module {
...
+   * Get the current owner of a given NFT within this Drop
+   *
+   * @param tokenId - the tokenId of the NFT
+   * @returns the address of the owner
+   */
+  public async ownerOf(tokenId: string): Promise<string> {
+    return await this.contractWrapper.readContract.ownerOf(tokenId);
+  }
+
+  // TODO extract all the claim condition logic to its own class
```
joaquim-verges, line 225: Are claim conditions reused in other modules? I'm gonna extract it out I think, it has a lot of logic, would be another namespace like `module.claimConditons.getActive()`

lmk what you think @atbe

src/modules/drop-erc-721.ts
```diff
@@ -56,7 +75,305 @@ export class DropErc721Module {
...
+  public async balance(): Promise<BigNumber> {
+    return await this.balanceOf(await this.contractWrapper.getSignerAddress());
+  }
+
+  /**
+   * Get whether this wallet has approved transfers from the given operator
+   * @param address - the wallet address
+   * @param operator - the operator address
+   */
+  public async isApproved(address: string, operator: string): Promise<boolean> {
```
joaquim-verges, line 320: how come we pass operator here? shouldn't it always be the drop contract address?

src/modules/drop-erc-721.ts
```diff
@@ -56,7 +75,305 @@ export class DropErc721Module {
...
+   * PRIVATE FUNCTIONS
+   *******************************/
+
+  private async getTokenMetadata(tokenId: string): Promise<NFTMetadata> {
+    // return await getTokenMetadata(
+    //   this.contractWrapper.readContract,
+    //   tokenId,
+    //   this.storage.ipfsGatewayUrl,
+    // );
+    // TODO common token metadata fetch
```
joaquim-verges, line 345: I didn't port this utility function yet - lmk what you want to do with this @jnsdls

src/modules/drop-erc-721.ts
```diff
@@ -56,7 +75,305 @@ export class DropErc721Module {
...
+    return Promise.resolve({});
+  }
+
+  private async transformResultToClaimCondition(
+    pm: IDropERC721.ClaimConditionStructOutput,
+  ): Promise<ClaimCondition> {
+    // TODO have a dedicated class for currency manipulation that takes in a contractWrapper?
+    // const cv = await getCurrencyValue(
```
joaquim-verges, line 353: same for this one - feels like all the currency stuff could also be encapsulated in a class

src/modules/drop-erc-721.ts
jnsdls: claim conditions exist on erc721 drop and erc1155drop (for now), they work slightly different in both cases but we might be able to extract them out

jnsdls:
let's keep the PR around for discussion but also move ahead and merge it and continue from there

PR #239, opened 2022-01-28
Test setup

atbe:


src/core/classes/factory.ts
```diff
@@ -15,7 +20,12 @@ export class ModuleFactory extends ContractWrapper<TWFactory> {
     storage: IStorage,
     options?: SDKOptions,
   ) {
-    super(network, TW_FACTORY_ADDRESS, TWFactory__factory.abi, options);
+    super(
+      network,
+      options?.thirdwebModuleFactory as string,
```
atbe, line 25: This will not always be passed, therefore we need to fetch it from the `CONTRACT_ADDRESSES` map

atbe:
Copied the deployment logic from here https://github.com/nftlabs/nftlabs-protocols/blob/v2/scripts/deploy/testDeployV2Setup.ts#L62

PR #247, opened 2022-02-03
Delayed reveal

jakeloo:
* rename isV1 -> isV0 to prevent confusion because that's old legacy version without `VERSION` in the contract. _version now tracks contract version.
* implements delay reveal
* tests for delay reveal
* tests for compatibility


src/modules/drop.ts
```diff
@@ -1122,52 +1173,229 @@ export class DropModule
...
+    );
+
+    return tokenUris
+      .map((uri, index) => ({
+        batchId: BigNumber.from(index),
+        batchUri: uri,
+        placeholderMetadata: tokenMetadatas[index],
+      }))
+      .filter(
+        (_, index) => ethers.utils.hexDataLength(encryptedBaseUris[index]) > 0,
```
joaquim-verges, line 1360: oh you changed this? was there an issue with the way i had it?

joaquim-verges (approved):
ship it

src/modules/drop.ts
jakeloo: yeah because bytes32 get parsed to hex data in ethers (i believe)
so what that means for empty encryptedBaseUris -- instead of "", it actually returned "0x". 
instead of doing string check on hex bytes, it's probably a better to use the proper function to check the length of the bytes.

Issue #242, opened 2022-02-01
updating claim wiping out snapshot


eabdelmoneim:
1. Create new drop module (not bundle drop)
2. Go to settings and upload a snapshot
3. Click save
4. Go back to the settings and change the "Price Per Token"
5. Click save

Suddenly the snapshot metadata is removed from the module metadata

atbe:
Fixed in #248

PR #248, opened 2022-02-03
Fix merkle wiped on updated claim conditions

atbe:
Fixes issue #242 

This bug **only effects the drop module**. It does not effect the Bundle Drop module because the Bundle Drop modules *appends* new root entries to the `merkle` dictionary, while the Drop module does an entire replace.


src/modules/drop.ts
```diff
@@ -19,6 +19,7 @@ import {
   ethers,
 } from "ethers";
 import { JsonConvert } from "json2typescript";
+import { C } from "ts-toolbelt";
```
jakeloo: is this used?

src/modules/drop.ts
```diff
@@ -594,13 +594,22 @@ export class DropModule
...
-    if (factory.allSnapshots().length === 0 && "merkle" in metadata) {
-      metadata["merkle"] = {};
-    } else {
+
+    const defaultMerkleRoot = hexZeroPad([0], 32).toString();
+
+    if (factory.allSnapshots().length > 0) {
       metadata["merkle"] = merkleInfo;
+    } else if (
+      factory.allSnapshots().length === 0 &&
```
jakeloo, line 607: this condition is always true :D

joaquim-verges (approved):
lg - i'll copy over your new tests in v2 to make sure i'm good there

src/modules/drop.ts
```diff
@@ -594,13 +594,22 @@ export class DropModule
...
+      conditions
+        .filter((c) => c.merkleRoot !== defaultMerkleRoot)
+        .map((c) => c.merkleRoot).length === 0
```
jakeloo, line 610: probably don't need `.map`

jakeloo (changes requested):

jakeloo (approved):

src/modules/drop.ts
atbe: been removed

src/modules/drop.ts
atbe: why do you think that?

src/modules/drop.ts
atbe: good point

src/modules/drop.ts
atbe: oh good point, the `if` condition right above 🤦🏽‍♂️

atbe: fixed

PR #251, opened 2022-02-05
bundle drop claim return transaction receipt

jarrodwatts:
Calling `bundleDropModule.claim` from the ts-sdk should return the transaction receipt as per the docs above the function.

Currently it returns void as nothing is being returned, hopefully this will fix that


joaquim-verges:
@jarrodwatts thank you for your contribution! We're going through a big refactor right now that returns transaction receipts for every transaction in a much nicer way than we have right now (checkout the v2 branch), we're very close so you'll get that very soon.

We could merge this now and make a new build for you (i'll ping you on discord when we get to it)

Alternatively you can get the transaction receipt right now by using the event handler in the sdk. something like:

```
sdk.event.on(EventType.Transaction, (event) => {
  const status = event.status; // will be either 'pending' or 'submitted'
  const transactionHash = event.transactionHash; 
}
```

would that be enough to unblock you for now?

PR #265, opened 2022-02-21
Use provider gas price and apply user options

joaquim-verges:
- remove gas station call, rely on provider gas price instead.
- Apply extra tip for transactions depending on the SDK options
- expose a `currentGasPriceInGwei()` convenience function


src/core/classes/contract-wrapper.ts
```diff
@@ -100,23 +99,39 @@ export class ContractWrapper<
...
-      speed,
-      maxGasPrice,
-    );
-    if (!gasPriceChain) {
-      return {};
-    }
-    // TODO: support EIP-1559 by try-catch, provider.getFeeData();
     return {
-      gasPrice: ethers.utils.parseUnits(gasPriceChain.toString(), "gwei"),
+      gasPrice: await this.getPreferredGasPrice(),
```
jakeloo, line 103: by setting `gasPrice` you'd always override the transaction type to be 1 instead of 2.

for setting transaction type 2, you'd need to do https://github.com/thirdweb-dev/contracts/blob/v2/scripts/deploy/testDeployV2Setup.ts#L43-L47

our sdk needs to work with both because fantom/avax are only supporting pre-eip1559.  polygon/eth are post-eip1559

joaquim-verges:
@jakeloo I made the change - I think this will work fine for mainnet. I'm a little worried about polygon though, they just added support for EIP1559 recently but I don't think ethers does the right thing there.

It hardcodes maxPriorityFeePerGas to 2.5 gwei no matter what, which feels scary.

this is what I get right now for polygon's `getFeeData()`:

```
priority fee 2.5 gwei
base fee 5.356588024 gwei
gasPrice 36.519404872 gwei
```

which means that if we do a 1559 type tx, by default we'd set the price to 2.5 + 5.35 = 7.87 gwei per gas unit, but the non-eip1559 gas price is 36 gwei (and i think the min gas price on polygon is 30 gwei IIRC)

We could have a minimum maxFeePerGas of 30 gwei i guess, but heistant to merge because of this, what do you think?

joaquim-verges:
@jakeloo also checking on metamask - they set a maxPriority fee of 33-36 gwei depending on the speed you want. Not sure how they calculate that, seems a bit high but probably more accurate than ethers' 2.5 gwei xD

jakeloo:
> @jakeloo I made the change - I think this will work fine for mainnet. I'm a little worried about polygon though, they just added support for EIP1559 recently but I don't think ethers does the right thing there.
> 
> It hardcodes maxPriorityFeePerGas to 2.5 gwei no matter what, which feels scary.
> 
> this is what I get right now for polygon's `getFeeData()`:
> 
> ```
> priority fee 2.5 gwei
> base fee 5.356588024 gwei
> gasPrice 36.519404872 gwei
> ```
> 
> which means that if we do a 1559 type tx, by default we'd set the price to 2.5 + 5.35 = 7.87 gwei per gas unit, but the non-eip1559 gas price is 36 gwei (and i think the min gas price on polygon is 30 gwei IIRC)
> 
> We could have a minimum maxFeePerGas of 30 gwei i guess, but heistant to merge because of this, what do you think?

yeah, this definitely seems broken. I think the reported fee data from the blockchain (polygon) doesn't account for the minimum gas price (30 gwei).  

i've definitely seen weird behavior for gas reporting for polygon. for example: 
image. 
notice how `Base: 0` and it's all priority fee.
ethereum mainnet for reference, there's a base fee + reasonable tips:
image



yeah generally the 2~2.5 gwei is sufficient for eth mainnet, but obviously it works a little differently than polygon due to minimum gas fee required for tx.

I think we should make sure that the priority fee is not lower than the minimum network fee (30 gwei for polygon). fyi, harmony chain has similar minimum fee model.

joaquim-verges:
@jakeloo there's also a new gasstation url for polygon https://gasstation-mainnet.matic.network/v2

i'm kinda worried about relying on ethers.js for polygon, what do you think?

joaquim-verges:
@jakeloo https://docs.polygon.technology/docs/develop/eip1559-transactions/how-to-send-eip1559-transactions/

jakeloo:
> @jakeloo there's also a new gasstation url for polygon https://gasstation-mainnet.matic.network/v2
> 
> i'm kinda worried about relying on ethers.js for polygon, what do you think?

agreed. ethers.js hardcoded it to 2.5 gwei, which is definitely only targeting ethereum chain. not toward polygon chain for the reason i mentioned above. 

we should keep our min 31 gwei, so we'll be at least 1 higher than the chain's minimum :)

src/core/classes/contract-wrapper.ts
```diff
@@ -105,21 +104,74 @@ export class ContractWrapper<
...
-    const speed = this.options.gasSettings?.speed || "fastest";
-    const maxGasPrice = this.options.gasSettings?.maxPriceInGwei || 300;
-    const gasPriceChain = await getGasPriceForChain(
-      chainId,
-      speed,
-      maxGasPrice,
-    );
-    if (!gasPriceChain) {
-      return {};
+    const feeData = await this.getProvider().getFeeData();
```
jakeloo, line 107: do you have to try catch this or nah? for chain that doesn't support eip1559.

PR #268, opened 2022-02-25
parse ethers.js errors and rethrow with human readable messages

joaquim-verges:
tries to detect ethers.js errors (which come with a 'reason' and 'code' field)
and extract what I believe is the most important piece of info : the 'message' field inside the body

example output:

```
Error: Contract transaction failed with message: "execution reverted: not enough minted tokens."

 reason: "cannot estimate gas; transaction may fail or may require manual gas limit"
 code: "UNPREDICTABLE_GAS_LIMIT"
 ```

this is where ethers.js creates their errors https://github.com/ethers-io/ethers.js/blob/73a46efea32c3f9a4833ed77896a216e3d3752a0/packages/logger/src.ts/index.ts#L189

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@thirdweb-dev/sdk",
-  "version": "2.0.0-nightly.25",
+  "version": "2.0.0-nightly.26",
   "description": "The main thirdweb SDK.",
   "repository": {
     "type": "git",
--- a/src/core/classes/contract-wrapper.ts
+++ b/src/core/classes/contract-wrapper.ts
@@ -200,7 +200,29 @@ export class ContractWrapper<
     if (!func) {
       throw new Error("invalid function");
     }
-    return await func(...args, callOverrides);
+    try {
+      return await func(...args, callOverrides);
+    } catch (e) {
+      throw this.readableErrorWithRevertReason(e);
+    }
+  }
+
+  private readableErrorWithRevertReason(e: any): Error {
+    if (e instanceof Error) {
+      const erasedError = e as any;
+      if (erasedError.reason && erasedError.code) {
+        // this is definitely a ethers.js error, try to extract error message in body
+        const regex = /.*message\\":\\"([a-zA-Z0-9 -_.]+)\\",.*/;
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by jakeloo** [comment]
I think the reason and code are misleading majority of the time. it's always "unpredictable gas".

I think the execution reverted message is the real error that should be highlighted. so that's good. 
one thought, i'm not sure how this work with custom error code: https://github.com/ethereum/EIPs/issues/838 

from debugging perspective, i think it's useful to show `from`, `to`, `data` because with these 3 information we can simulate the transaction and figure out what issues people are running into (if they ask for support).

**Comment by joaquim-verges** [comment]
yeah i wasn't sure if reason/code ever have something interesting - I want to make sure I don't hide information that we'd need to debug stuff.

Im gonna add `from` `to` and `chainId`

@jakeloo 2 questions:
- is it possible for me to decode the `data` field somehow using ethers?
- how do you feel about having the RPC url in the error msg as well? (might expose some keys, don't want ppl to copy pasta it in discord, but could be useful to see that you put in a wrong RPC url)

**Comment by joaquim-verges** [comment]
@jakeloo this is the new format:

```
Error: Contract transaction failed: "execution reverted: not enough minted tokens."

Transaction info:
- from: "0xe79ee09bd47f4f5381dbbacacff2040f2fbc5803"
- to: "0x0e82b49a9a010ea4abbcd4e5e0d115a59bf1cc00"
- chain: "maticmum" (80001)
- data: "0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000196e6f7420656e6f756768206d696e74656420746f6b656e732e00000000000000"
```

i think this is good for now - gonna merge

jakeloo:
> * is it possible for me to decode the `data` field somehow using ethers?

yes, you could do `contract.interface.decodeFunctionData`.

> how do you feel about having the RPC url in the error msg as well? (might expose some keys, don't want ppl to copy pasta it in discord, but could be useful to see that you put in a wrong RPC url)

one idea would be just include the host name -- without path or query params

PR #285, opened 2022-03-11
Enable custom trusted forwarders + always deploy with OZ defender + biconomy setup

joaquim-verges:


jakeloo:
image

we should do sanity check on the currency input. @joaquim-verges SDK should prevent this from happening. we should be able to check if the token input supports delegation / voting or not.

jakeloo:
do we have preference on `trusted_forwarder` vs `trusted_forwarders`?

joaquim-verges:
@jakeloo done

jakeloo (approved):

jakeloo:
actually 2 things:
1. getDefaultTrustedForwarders() doesn't seem to handle for AddressZero if biconomy not supported on the chain?
2. i do think there's a benefit for having full override to remove trusted forwarders from the default. what do you think? maybe we can add it in the future to do SET rather than CONCAT.

joaquim-verges:
What happens if I set address zero as the forwarder ? Assumed it was a no-op

Set vs concat - what's the harm in having the default ones in there as well? That way you could switch between your custom one or the default one if you want to?

jakeloo:
1. it's not truly noop, it'd add address zero as the trusted forwarder.
addresses in the trusted forwarder would means that they can override transactions `msg.sender`, i'd much rather keep tight and concise.
2. for the same reason above, given how serious the serious risk is, we should give out option that's more concise. if i don't trust biconomy's trusted forwarder contract -- i should be able to get rid of it and only put in my own custom contract.

joaquim-verges:
@jakeloo fixed

PR #286, opened 2022-03-12
remove local docusaurus

joaquim-verges:
not needed anymore now that we have the docs repo setup


joaquim-verges:
@jnsdls this should be safe right? just disable the job on zeet and that's it? https://zeet.co/nftlabs/nftlabs-typescript-sdk

PR #306, opened 2022-03-19
fix claimeligibility reasons

jnsdls:
v1 contracts did not expose the necessary function for this so we hacked around it with balanceOf
in v2 we can do the check properly


joaquim-verges (approved):
Lg

PR #310, opened 2022-03-22
Add functions for buildspace

nachoiacovino:


src/core/classes/drop-erc1155-history.ts
```diff
@@ -0,0 +1,36 @@
+import { DropERC1155 } from "@thirdweb-dev/contracts";
+import { ContractWrapper } from "./contract-wrapper";
+import { BigNumberish } from "ethers";
+
+/**
+ * Manages claim conditions for Edition Drop contracts
```
joaquim-verges, line 6: update doc?

src/core/classes/drop-erc1155-history.ts
```diff
@@ -0,0 +1,36 @@
...
+   * @beta - This can be very slow for large numbers of token holders
+   *
+   * @param tokenId - The token id to get the claimers of
+   * @returns - A unique list of addresses that claimed the token
+   */
+  public async getAllClaimerAddresses(
+    tokenId: BigNumberish,
+  ): Promise<string[]> {
+    const a = await this.contractWrapper.readContract.queryFilter(
+      this.contractWrapper.readContract.filters.TokensClaimed(null, tokenId),
```
joaquim-verges, line 32: btw the reason why i removed these is because this will get limited to a few hundred/thousand blocks depending on the RPC you're using. The hope was to replace this with the graph, but doesn't look like it's happening any time soon right @jakeloo ?

src/core/classes/erc-20-history.ts
```diff
@@ -0,0 +1,55 @@
+import { ContractWrapper } from "./contract-wrapper";
+import { BigNumber } from "ethers";
+import { TokenERC20 } from "@thirdweb-dev/contracts";
+import { AddressZero } from "@ethersproject/constants";
+
+/**
+ * Manages claim conditions for Edition Drop contracts
```
joaquim-verges, line 7: update comment

joaquim-verges (approved):
just minor comments, would be really great to add a quick unit test for each function (should be really easy to add to an existing test for example) but I can do it for you when i get back and add more of these.

feel free to merge to unblock you though

src/core/classes/drop-erc1155-history.ts
nachoiacovino: jake mentioned that's still gonna be the case so we're just adding these two functions to unblock buildspace and then we can start thinking how to do this with the graph going forward

jakeloo: subgraph is probably a couple weeks away. we should unblock buildspace first, then we'll swap out the implementation later.

PR #313, opened 2022-03-22
Fix transferBatch on token contract

nachoiacovino:
Also Convert to BigNumber to be able to pass tokenId as string on getAllClaimerAddresses

### DIFF
--- a/src/core/classes/erc-20.ts
+++ b/src/core/classes/erc-20.ts
@@ -310,12 +310,17 @@ export class Erc20<T extends TokenERC20> implements UpdateableNetwork {
    * ```
    */
   public async transferBatch(args: TokenMintInput[]) {
-    const encoded = args.map((arg) =>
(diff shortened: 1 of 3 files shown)


src/core/classes/erc-20.ts
```diff
@@ -310,12 +310,17 @@ export class Erc20<T extends TokenERC20> implements UpdateableNetwork {
    * ```
    */
   public async transferBatch(args: TokenMintInput[]) {
-    const encoded = args.map((arg) =>
-      this.contractWrapper.readContract.interface.encodeFunctionData(
+    const decimals = await this.contractWrapper.readContract.decimals();
```
joaquim-verges, line 313: There's a helper function for this already, look for `normalizePrice`

joaquim-verges: actually nevermind - this is fine here because we don't want to re-fetch the token info

joaquim-verges (approved):

PR #316, opened 2022-03-24
naive custom contract init

jnsdls:
enables: https://github.com/thirdweb-dev/react/pull/13, https://github.com/thirdweb-dev/dashboard/pull/20


src/contracts/custom.ts
```diff
@@ -0,0 +1,61 @@
...
+  public metadata;
+  private storage;
+
+  constructor(
+    network: NetworkOrSignerOrProvider,
+    address: string,
+    storage: IStorage,
+    options: SDKOptions = {},
+    abi: any = undefined,
+    contractWrapper = new ContractWrapper<IThirdwebContract>(
```
joaquim-verges, line 30: how about typing the `ContractWrapper` with `BaseContract` instead? that way we don't even require `IThirdwebContract`, should work with any contract and you dont need to do that weird try catch in the sdk.get function?

src/contracts/pack.ts
```diff
@@ -131,7 +131,7 @@ export class Pack implements UpdateableNetwork {
    * @param packId - the id of the pack to fetch
    * @returns the pack metadata
    */
-  public async get(packId: string): Promise<PackMetadata> {
+  public async get(packId: BigNumberish): Promise<PackMetadata> {
```
joaquim-verges, line 134: let's make this change separately no? seems like a bugfix we should ship independently

src/contracts/custom.ts
```diff
@@ -0,0 +1,61 @@
...
+      network,
+      address,
+      abi || IThirdwebContract__factory.abi,
+      options,
+    ),
+  ) {
+    this.storage = storage;
+    this.contractWrapper = contractWrapper;
+    this.metadata = new ContractMetadata(
```
joaquim-verges, line 39: gonna have to figure out the flow of 
- detect interface
- instantiate corresponding 'feature' class

and apply that flow for all of these

src/contracts/custom.ts
jnsdls: yup fair enough, just means we don't get *anything* on it by default, we need to feature detect every thing (but we already need to do that anyways so I think that's probably fine)

src/contracts/custom.ts
jnsdls: yup, will spend a bit of time on this today and improve it in this branch / get to something that could work

src/contracts/pack.ts
jnsdls: https://github.com/thirdweb-dev/typescript-sdk/pull/317

src/contracts/custom.ts
jnsdls: yeah, give it another look, let me know if it's directionally more like what ur thinking, then I can unblock /react from it and u can take it over tomorrow from there?

PR #320, opened 2022-03-27
implement erc20 signature minting

joaquim-verges:
(WIP)


joaquim-verges:
@jakeloo @nkrishang I'm running into issues verifying the erc20 signatures, I triple checked everything, but can't see what's wrong. erc721/1155 work with the same setup. Can you see anything I'm doing wrong?

Here's an example payload I'm signing:
```
{
  to: '0x70997970C51812dc3A010C7d01b50e0d17dc79C8',
  primarySaleRecipient: '0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
  quantity: BigNumber { value: "50000000000000000000" },
  price: BigNumber { value: "200000000000000000" },
  currency: '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee',
  validityStartTimestamp: BigNumber { value: "1648441503" },
  validityEndTimestamp: BigNumber { value: "1963801503" },
  uid: '0x3936336333663838306263653464623862303833313338333235353336613931'
}
```

nkrishang:
I see this test case is failing [`should generate a valid signature`](https://github.com/thirdweb-dev/typescript-sdk/blob/joaquim/sigmint_erc20/test/signature-mint-20.test.ts#L76)

The `verify` function will return `true` given two conditions (1) the UID of the payload should not have already been used on the contract, and (2) the account that signed the payload must have minter role at the time of calling `verify`. 

In the SDK test case, can we check which of the two is failing? How to check: `verify` returns the `signer` i.e. the account that signed off on the payload. If this `signer` has `MINTER_ROLE` on the contract (`contract.hasRole(signer, MINTER_ROLE)`) then the check (1) is failing, else (2) is failing.

I'm still looking; up till now, haven't found any difference in the signature handling code between `TokenERC721` and `TokenERC20`.

joaquim-verges:
Both mint and verify fail at the same point when trying to recover the address from a mint request + signature.

My latest hunch from debugging with tenderly is that my mint request object is not transmitted properly between clone and implementation 🤔

joaquim-verges:
@nkrishang do we have a test for sigmint erc20 in contracts?

PR #322, opened 2022-03-28
cache the resolve type

jnsdls:


jnsdls:
cancelling this in favor of BYOC

PR #324, opened 2022-03-29
Introduce Transaction Interceptor concept 

joaquim-verges:
allows to override the parameters of the next executed transaction

API:
```
contract.interceptor.overrideNextTransaction(() => ({
      nonce: 123,
    }));
// next write call will augment CallOverrides with the interceptor data
      ```

### CODE REVIEW

**Review by jakeloo — APPROVED** [review]

File: src/contracts/edition-drop.ts [hunk 1]
```diff
@@ -107,6 +108,7 @@ export class EditionDrop extends Erc1155<DropERC1155> {
    */
   public claimConditions: DropErc1155ClaimConditions;
   public history: DropErc1155History;
+  public interceptor: TransactionInterceptor<DropERC1155>;
```
jakeloo, line 111: probably wanna `@internal` here?

src/contracts/nft-drop.ts
```diff
@@ -61,6 +62,7 @@ export class NFTDrop extends Erc721<DropERC721> {
   public metadata: ContractMetadata<DropERC721, typeof NFTDrop.schema>;
   public primarySale: ContractPrimarySale<DropERC721>;
   public events: ContractEvents<DropERC721>;
+  public interceptor: TransactionInterceptor<DropERC721>;
```
jakeloo, line 65: missing `@internal`?

PR #327, opened 2022-03-31
Token Drop Contract

joaquim-verges:


src/contracts/nft-drop.ts
```diff
@@ -109,7 +109,7 @@ export class NFTDrop extends Erc721<DropERC721> {
    * await dropContract.claimConditions.set(claimConditions);
    * ```
    */
-  public claimConditions: DropErc721ClaimConditions;
+  public claimConditions: DropErc721ClaimConditions<DropERC721>;
```
jnsdls, line 112: since this is a generic type now do we still need the specific `DropErc721ClaimConditions` name? or can it become a generic `ClaimConditions<DropErc721>` ?

src/contracts/token-drop.ts
```diff
@@ -0,0 +1,203 @@
...
+  static schema = DropErc20ContractSchema;
+
+  public metadata: ContractMetadata<DropERC20, typeof TokenDrop.schema>;
+  public roles: ContractRoles<
+    DropERC20,
+    typeof TokenDrop.contractRoles[number]
+  >;
+  public encoder: ContractEncoder<DropERC20>;
+  public estimator: GasCostEstimator<DropERC20>;
+  public claimConditions: DropErc721ClaimConditions<DropERC20>;
```
jnsdls, line 54: same question, this reads really odd, which one is it 721 or 20?

src/core/classes/erc-20.ts
```diff
@@ -396,4 +376,9 @@ export class Erc20<T extends TokenERC20> implements UpdateableNetwork {
       BigNumber.from(value),
     );
   }
+
+  protected async normalizeAmount(amount: Amount): Promise<BigNumber> {
+    const decimals = await this.contractWrapper.readContract.decimals();
```
jnsdls, line 381: maybe over-optimization but I wonder if we can cache this on the class itself (aka after the first time we look it up we just know what it is. decimal setting is not changeable so once we have it for a given contract instance once we should be able to just return it immediately instead of calling the contract)

jnsdls (commented):
left some notes

src/contracts/nft-drop.ts
joaquim-verges: Yep, just hadn't renamed yet. Still some things to work out in this pr

src/core/classes/erc-20.ts
joaquim-verges: That's a good point. Doesn't hurt to cache this

joaquim-verges:
Went down a rabbit whole to support passing token quantities in eth (instead of wei) in the claim conditions for things like max quantity and max claimable per wallet.

I'm close but it since I'm sharing API for erc20 and erc721 it got way more involved than I thought. 

Will prob separate it in another pr

PR #329, opened 2022-04-01
add resolveContract and contractType resolver caching

jnsdls:


jnsdls:
upgrading deps added a bunch of changes to MD files on build... whoops :)

PR #334, opened 2022-04-03
update ethers to 5.6.2 and fix error parsing

joaquim-verges:


package.json
```diff
@@ -74,7 +74,7 @@
     "typescript": "^4.6.2"
   },
   "peerDependencies": {
-    "ethers": ">=5 <5.6.2"
+    "ethers": "^5"
   },
   "dependencies": {
     "@ethersproject/bignumber": "5.6.0",
```
jnsdls, line 80: probably need to revert these to ^ notation again as well

PR #341, opened 2022-04-10
fix tokenId filter comparison on getActiveListings

jarrodwatts:


joaquim-verges (approved):
Good catch! Thanks @jarrodwatts

joaquim-verges:
Will merge when I get to a computer, add a couple of tests and release. Thanks again!

PR #344, opened 2022-04-15
fix resetting snapshots when updating claim conditions

joaquim-verges:


joaquim-verges:
@jnsdls was probably the reason why we were getting mismatching merkle roots

jnsdls:
Oh nice, this is a good find!

jnsdls (approved):
LGTM

PR #348, opened 2022-04-16
Expose transaction events for each contract

joaquim-verges:
new API for listening to transaction status on each contract

```
contract.events.addTransactionListener((event) => {
      console.log(event);
});
 ```


jnsdls (approved):
LGTM

PR #357, opened 2022-04-26
Add react examples

adam-maj:


joaquim-verges:
the snippets already have the concept of "language", would be nice to pull them into 3 different tabs in the dashboard: JS / React / Python and switch the entire list of code snippets. Should be able to fetch that snippet file from 3 diff repos rather than having it all in here right?

PR #366, opened 2022-05-01
fix provider only & promise things

jnsdls:


src/core/classes/contract-publisher.ts
```diff
@@ -113,7 +113,8 @@ export class ContractPublisher extends RPCConnectionHandler {
     const contract = new Contract(
       address,
       ThirdwebContract__factory.abi,
-      this.getSignerOrProvider(),
```
jnsdls, line 116: @joaquim-verges this was the actual cuplrit

PR #378, opened 2022-05-06
Add ERC-20 airdrop

nachoiacovino:


joaquim-verges:
closing this for now - we have transfer batch if ppl want to do token airdrops. We'll make this better as we modularize this into a contract extension

PR #382, opened 2022-05-09
filebase

adam-maj:


src/core/classes/filebase-storage.ts
```diff
@@ -0,0 +1,282 @@
...
+}
+
+/**
+ * IPFS Storage implementation, accepts custom IPFS gateways
+ * @public
+ */
+export class IpfsStorage implements IStorage {
+  private gatewayUrl: string;
+
+  constructor(gatewayUrl: string = DEFAULT_IPFS_GATEWAY) {
```
joaquim-verges, line 43: what would be neat here is to pass a network abstraction here that just does the "fetch" and the "upload" - the rest of the logic can be completely shared between different IPFS providers

jnsdls:
stale

PR #396, opened 2022-05-12
Implement Signature Drop SDK wrapper

ndeto:
[Linear Ticket](https://linear.app/thirdweb/issue/SDK-279/implement-signaturedrop-sdk-wrapper)

This PR implements the SignatureDrop SDK wrapper

Context:

I've named it `MyNFTDrop` because of the conflicts I was getting while using `NFTDrop` . I'd appreciate suggestions.
The documentation is not accurate as well. To be corrected
As for the tests, I'm running the `yash/sign-drop-test` branch for the smart contracts since there's a different implementation with the permissions.

I'd love comments on any of the above, as well as feedback on the general things i might have missed while creating the MyNFTClass.


src/core/classes/erc-721-with-quantity-signature-minting.ts
```diff
@@ -0,0 +1,288 @@
...
+   * };
+   *
+   * const signedPayload = contract.signature.generate(payload);
+   * // now anyone can use these to mint the NFT using `contract.signature.mint(signedPayload)`
+   * ```
+   * @param mintRequest - the payload to sign
+   * @returns the signed payload and the corresponding signature
+   */
+  public async generate(
+    mintRequest: PayloadToSign1155 | PayloadToSign721withQuantity,
```
joaquim-verges, line 196: remove 1155 inputs from here and below

src/core/classes/factory.ts
```diff
@@ -120,6 +122,35 @@ export class ContractFactory extends ContractWrapper<TWFactory> {
...
+        const signer = this.getSigner();
+        invariant(signer);
+        const sigMintDeployer = await new ethers.ContractFactory(
+          SigMint__factory.abi,
+          SigMint__factory.bytecode,
+        )
+          .connect(signer)
+          .deploy();
+
+        await sigMintDeployer.deployTransaction.wait();
```
joaquim-verges, line 136: @nkrishang this requires users to do 2 transactions when deploying a SignatureDrop contract from the dashboard/sdk. I think we can avoid this by just pre-deploying the `SigMint` contract on every chain. It should be ok to use the `SigMint` same contract for any number of `SignatureDrop` contracts right?

src/core/classes/erc-721-with-quantity-signature-minting.ts
```diff
@@ -0,0 +1,288 @@
+import {
+  FilledSignature721WithQuantity,
+  MintRequest721withQuantity,
+  PayloadToSign1155,
+  PayloadToSign721withQuantity,
+  PayloadWithUri721withQuantity,
+  Signature721WithQuantityInput,
+  Signature721WithQuantityOutput,
+  SignedPayload1155,
```
joaquim-verges, line 9: remove unused imports

src/core/classes/erc-721-with-quantity-signature-minting.ts
```diff
@@ -0,0 +1,288 @@
...
+} from "contracts";
+import { IStorage } from "../interfaces";
+import { ContractRoles } from "./contract-roles";
+import { SignatureDrop } from "../../contracts";
+import { BigNumber } from "ethers";
+import { uploadOrExtractURIs } from "../../common/nft";
+import { TokensMintedWithSignatureEvent } from "contracts/SigMint";
+
+/**
+ * Enables generating dynamic ERC1155 NFTs with rules and an associated signature, which can then be minted by anyone securely
```
joaquim-verges, line 28: update documentation

src/core/classes/erc-721-with-quantity-signature-minting.ts
joaquim-verges: still have an 1155 input here

PR #397, opened 2022-05-13
TokenDrop deploy metadata

nachoiacovino:


jnsdls:
hm, tokendrop actually *does* support platform fee...

PR #414, opened 2022-05-21
removed extra character in javascript code snippet

robriks:
removed an extra / in js code snippet my_script.js (looked like it was a remnant of a comment that was removed at some point)

cheers <3


jnsdls:
thanks @robriks!

PR #422, opened 2022-05-27
Add example js snippets for deploying pre-built contracts

jarrodwatts:


src/core/classes/contract-deployer.ts
```diff
@@ -162,6 +252,22 @@ export class ContractDeployer extends RPCConnectionHandler {
...
+   *
+   * @example
+   * ```javascript
+   * const contractAddress = await sdk.deployer.deploySplit({
+   *   name: "My Split",
+   *   primary_sale_recipient: "your-address",
+   *   recipients: [
+   *    {
+   *      address: "your-address",
+   *      sharesBps: 100 * 100, // 100%
```
joaquim-verges, line 266: Would be nice to show a 2 person split here like a 80/20

joaquim-verges (approved):
Very nice - thanks for adding these!

src/core/classes/contract-deployer.ts
jarrodwatts: added

PR #425, opened 2022-05-28
Add decimals to getWrappedContents and fix example

adam-maj:


joaquim-verges (approved):
Thanks for spotting it!

PR #428, opened 2022-05-31
Allow for explicit entry of bytes as bytes32 param when publishing

bagelface:
Previously the deployment form would throw "bytes32 string must be less than 32 bytes" (see screenshot) when entering 32bytes directly as a byte string. This is because `formatBytes32String` was being used which is for converting strings to bytes. Instead, should be using `hexlify` and `hexZeroPad` functions

Screen Shot 2022-05-31 at 9 50 10 AM

### DIFF
--- a/src/core/classes/contract-publisher.ts
+++ b/src/core/classes/contract-publisher.ts
@@ -284,7 +284,7 @@ export class ContractPublisher extends RPCConnectionHandler {
         }
       }
       if (p === "bytes32") {
-        if (!isNaN(constructorParamValues[index])) {
+        if (!isNaN(constructorParamValues[index]) && constructorParamValues[index] > 0) {
           return ethers.utils.hexZeroPad(
             ethers.utils.hexlify(
               ethers.BigNumber.from(constructorParamValues[index]),
--- a/src/core/classes/contract-publisher.ts
+++ b/src/core/classes/contract-publisher.ts
@@ -284,7 +284,7 @@ export class ContractPublisher extends RPCConnectionHandler {
         }
       }
       if (p === "bytes32") {
-        if (!isNaN(constructorParamValues[index]) && constructorParamValues[index] > 0) {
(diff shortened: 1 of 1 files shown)


src/core/classes/contract-publisher.ts
```diff
@@ -284,9 +284,18 @@ export class ContractPublisher extends RPCConnectionHandler {
         }
       }
       if (p === "bytes32") {
-        return ethers.utils.formatBytes32String(
-          constructorParamValues[index].toString(),
-        );
+        if (!isNaN(constructorParamValues[index]) && constructorParamValues[index] >= 0) {
```
joaquim-verges, line 287: `constructorParamValues[index]` will always be a string - this check won't work.

Howw about `ethers.utils.isHexString(constructorParamValues[index])` instead (which means you don't have to hexlify under the hood, just pad)

joaquim-verges (changes requested):

src/core/classes/contract-publisher.ts
bagelface: `isNaN('0x1234')` returns `false`
`isNaN('1234')` returns `false`

joaquim-verges: @bagelface we use typescript, and in typescript `isNan()` expects a number, so it throws a type error (won't compile)

joaquim-verges: but in that case their contract param type won't be `bytes32` right? would probably be `uint256` or something for a number parameter.

bagelface: Lol seems a little counter-intuitive to only accept a number for a "is not a number" check. We can wrap `constructorParamValues[index]` as `Number(constructorParamValues[index])` if you're ok with my previous comment about "1234" being treated like an integer

joaquim-verges:
thanks again for the contribution! closing to merge the other PR

PR #430, opened 2022-05-31
split node and browser builds and export them separately to avoid dependency issues

jnsdls:


package.json
```diff
@@ -46,17 +57,14 @@
...
     "@types/deep-equal-in-any-order": "^1.0.1",
     "@types/expect": "^24.3.0",
-    "@types/form-data": "^2.5.0",
     "@types/mocha": "^9.1.1",
     "@types/node": "^17.0.0",
     "@types/uuid": "^8.3.3",
     "@typescript-eslint/eslint-plugin": "^5.2.0",
     "@typescript-eslint/parser": "^5.2.0",
     "chai": "^4.3.6",
-    "deep-equal": "^2.0.5",
```
jnsdls, line 59: was build a pain to work with, replaced with `fast-deep-equal`

tsup.config.ts
```diff
@@ -19,8 +19,33 @@ export default defineConfig([
     treeshake: true,
     globalName: "ThirdwebSDK",
     format: ["cjs", "esm"],
+    outDir: "dist/node",
+    banner: {
+      js: `import "cross-fetch/dist/node-polyfill.js";`,
```
jnsdls, line 24: importing the cress-fetch polyfill only in the node case (this is injected onto the top of the file)

PR #432, opened 2022-06-01
Stricter bytes and bytes32 inputs as hex strings only

joaquim-verges:


src/core/classes/contract-publisher.ts
```diff
@@ -284,14 +284,20 @@ export class ContractPublisher extends RPCConnectionHandler {
         }
       }
       if (p === "bytes32") {
-        return ethers.utils.formatBytes32String(
-          constructorParamValues[index].toString(),
+        invariant(
+          ethers.utils.isHexString(constructorParamValues[index]),
+          `Could not parse bytes32 value. Expected valid hex string but got "${constructorParamValues[index]}".`,
         );
+        if (ethers.utils.isHexString(constructorParamValues[index])) {
```
bagelface, line 291: this seems redundant because of the `invariant` function before it. if there's a type issue probably better to just use `constructorParamValues[index] as string`

joaquim-verges: woops - left it on accident. thank you for catching it :P

PR #438, opened 2022-06-01
Use deep-equal again

nachoiacovino:


nachoiacovino:
not working yet, getting a different error

nachoiacovino:
nope

PR #440, opened 2022-06-01
Remove un-used byoc registry and factory

joaquim-verges:


src/core/sdk.ts
```diff
@@ -121,7 +120,7 @@ export class ThirdwebSDK extends RPCConnectionHandler {
    * @internal
    * should never be accessed directly, use {@link ThirdwebSDK.getPublisher} instead
    */
-  private _publisher: Promise<ContractPublisher> | undefined;
+  private _publisher: ContractPublisher;
```
jnsdls, line 123: nice that we can make it sync again

PR #442, opened 2022-06-02
[fix] handle the additional usage of Buffer in browser builds

jnsdls:


joaquim-verges (approved):
Starting to not like that ts-up config file lol

PR #444, opened 2022-06-02
Bugfix - Incorrect entrypoint being loaded when using Node ESModules

jameshowe:
When attempting to load the SDK from a Node script the Browser package is being loaded. This is because `module` does not appear to be a [known condition](https://nodejs.org/api/packages.html#conditional-exports) and therefore it skips and loads the `default` entrypoint (`browser`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "unpkg": "dist/browser/index.global.js",
   "exports": {
     "node": {
-      "module": "./dist/node/index.mjs",
+      "import": "./dist/node/index.mjs",
       "require": "./dist/node/index.js"
     },
     "default": "./dist/browser/index.mjs"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
   "unpkg": "dist/browser/index.global.js",
   "exports": {
     "node": {
-      "module": "./dist/node/index.mjs",
+      "import": "./dist/node/index.mjs",
```
jnsdls, line 17:
```suggestion
      "import": "./dist/node/index.mjs",
      "module": "./dist/node/index.mjs",
```

we want to keep `module` around in addition (there are some tools and bundlers that can pick up `module` but not `import`)

jameshowe: No worries, wasn't aware, done :)

jnsdls (approved):

jnsdls:
thanks @jameshowe!

jameshowe:
> thanks @jameshowe!

No worries! Any view on when this will be released? (I can patch my local version just now)

PR #447, opened 2022-06-02
Fix get all NFTs bug on multiwrap

adam-maj:


jnsdls:
pretty significant change in behavior (just skip past failed tokens?)

also means we cannot rely on pagination anymore

src/core/classes/erc-721-supply.ts
```diff
@@ -65,11 +65,18 @@ export class Erc721Supply implements DetectableFeature {
...
-    return await Promise.all(
-      [...Array(maxId - start).keys()].map((i) =>
-        this.erc721.get((start + i).toString()),
-      ),
-    );
+
+    const nfts = [];
+    for (let i = start; i < maxId; i++) {
+      try {
+        const nft = await this.erc721.get(i);
```
joaquim-verges, line 72: this makes it all sequential which is super slow...

what's the problem you're trying solve here? just not throw on not found ?

joaquim-verges (changes requested):
requires a change in the contract - the burn behavior is different than the other NFT contracts (should preserve URI and just set owner to address zero)

joaquim-verges:
closing this - just need to deploy the changes in the contract

PR #450, opened 2022-06-03
Add support for wallet authenticator

adam-maj:


src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,184 @@
...
+  AuthorizedPayload,
+  AuthorizedPayloadSchema,
+  FilledAuthenticationPayloadInput,
+} from "../../schema/auth";
+import { RPCConnectionHandler } from "../classes/rpc-connection-handler";
+import { NetworkOrSignerOrProvider } from "../types";
+
+/**
+ * Wallet Authenticator
+ * @remarks Enables sign-in with ethereum based authentication.
```
joaquim-verges, line 19: could expand a bit more on the fact that this is using a JWT style authentication system, token can be re-used for multiple calls, and does not need to be stored anywhere

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,184 @@
...
+      authenticatedPayload,
+    });
+
+    return payload;
+  }
+
+  public async verify(
+    application: string,
+    authenticatedPayload: AuthenticatedPayload,
+  ): Promise<boolean> {
```
joaquim-verges, line 115: wonder if this should return 'why' verification failed. maybe with an enum?

src/schema/auth.ts
```diff
@@ -0,0 +1,130 @@
...
+   */
+  application: z.string(),
+  /**
+   * The address of the wallet requesting to authenticate
+   */
+  subject: AddressSchema,
+  /**
+   * The audience of endpoints intended to receive the authentication payload
+   */
+  audience: z.array(z.string()).optional(),
```
joaquim-verges, line 20: should we just call this endpoints?

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,184 @@
...
+      signerAddress.toLowerCase()
+    ) {
+      throw Error(
+        "Payload subject is not the same as the wallet connected to the SDK.",
+      );
+    }
+
+    const message = this.generateMessage(authorizedPayload);
+    const signer = this.requireSigner();
+    const authenticatedPayload = await signer.signMessage(message);
```
joaquim-verges, line 103: how about using the `UserWallet` class here? I expose a bunch of things that you're re-implementing here like `requireSigner` and `signMessage`

you could pass in a `UserWallet` class to the constructor of this class

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,189 @@
...
+ *   subject: "0x...",
+ * })
+ *
+ * // Then on the client side, we can sign this payload with the wallet requesting to authenticate
+ * const authenticatedPayload = await sdk.auth.sign(payloadWithSignature)
+ *
+ * // Finally, on the server side, we can verify if this payload is valid
+ * const isValid = await sdk.auth.verify(application, authenticatedPayload)
+ * ```
+ * @public
```
joaquim-verges, line 37: let's mark this @beta for now. We need to make sure this is rock solid with no possible exploit before we go full prod

src/core/auth/wallet-authenticator.ts
adam-maj: It would need to be stored on the client-side if they want to be able to make multiple requests w/o signing in again - but the nice thing is doesn't need any storage in a database

src/schema/auth.ts
adam-maj: Hmm - I do think this field could use a rename potentially (although I think deferring to the JWT is a good default if we have nothing else) - but I think audience may be too limiting. The intention of the `aud` field on JWT, although it's often used to specify which endpoints a token can auth to, isn't always used like that.

In fact in the JWT spec, audience is supposed to be a field that's open to whatever use the backend wants to use it for.

src/core/auth/wallet-authenticator.ts
adam-maj: I don't know if we want verification to return something besides a true/false value here - reason being that the main concern of the endpoint is usually just to check if the user _is_ or _isnt_ authed, and are more concerned with the actual endpoint functionality - the authentication should just be a simple check.

Feels like returning something like `getClaimIneligibilityReasons` with an array of reasons it failed/single reason it failed could be making this common use less intuitive (no longer checking if `authenticated == true`) + with drop it makes more sense b/c there will commonly be times when users don't meet the conditions to claim - whereas in authentication, this verify would really only fail if the user is an attacker, and has tampered with the payload.

If we people specifically need the reasons that a verification failed, maybe it would be good to expose a separate function that explicitly returns reasons for failure which can be used for that case, and maybe internally w/ the verify function

src/core/auth/wallet-authenticator.ts
adam-maj: good idea

src/core/auth/wallet-authenticator.ts
adam-maj: Although, as we add in checks for correct application/endpoint, maybe this would be helpful for debugging those 2 cases

src/schema/auth.ts
adam-maj: Ill change it to `endpoints` for now for simplicity - if we need to expand scope later we can

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,342 @@
...
+   */
+  public async login(
+    domain: string,
+    options?: LoginOptions,
+  ): Promise<LoginPayload> {
+    const parsedOptions = LoginOptionsSchema.parse(options);
+
+    const signerAddress = await this.wallet.getAddress();
+    const expirationTime =
+      parsedOptions?.expirationTime || new Date(Date.now() + 1000 * 60 * 5);
```
joaquim-verges, line 85: 5 minutes by default? is that from the standard?

btw you could put this in the zod schema if you want to

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,342 @@
...
+   * const address = sdk.auth.verify(domain, loginPayload);
+   * ```
+   */
+  public verify(
+    domain: string,
+    payload: LoginPayload,
+    options?: VerifyOptions,
+  ): string {
+    if (isBrowser()) {
+      throw new Error("Should not verify login on the browser.");
```
joaquim-verges, line 127: wonder if we should explain why here?

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    const parsedOptions = VerifyOptionsSchema.parse(options);
+
+    // Check that the intended domain matches the domain of the payload
+    if (payload.payload.domain !== domain) {
+      throw new Error(
+        `Expected domain ${domain} does not match domain on payload ${payload.payload.domain}`,
```
joaquim-verges, line 135: nit: little quotes around the values would be clutch here and below

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,342 @@
...
+    const userAddress = this.verify(domain, payload);
+    const adminAddress = await this.wallet.getAddress();
+    const payloadData = AuthenticationPayloadDataSchema.parse({
+      iss: adminAddress,
+      sub: userAddress,
+      aud: domain,
+      nbf: parsedOptions?.invalidBefore || new Date(),
+      exp:
+        parsedOptions?.expirationTime ||
+        new Date(Date.now() + 1000 * 60 * 60 * 5),
```
joaquim-verges, line 208: same question here, where does the 5h come from? Unsure what i would expect as a user if i don't pass expiration time, my first reaction would be - don't expire it. but maybe that's dangerous?

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,342 @@
...
+    const encodedSignature = Buffer.from(signature).toString("base64");
+
+    // Generate a JWT token with base64 encoded header, payload, and signature
+    const token = `${encodedHeader}.${encodedData}.${encodedSignature}`;
+
+    return token;
+  }
+
+  /**
+   * Authenticate With Token
```
joaquim-verges, line 237: wonder if we should specify 'auth token' here and below, since we use token for many other things in the sdk.

same for `generate` - think we can be more specific, something like `generateAuthToken`

src/core/auth/wallet-authenticator.ts
```diff
@@ -0,0 +1,342 @@
...
+   *
+   * @example
+   * ```javascript
+   * // Add the domain of the application users will login to, this will be used throughout the login process
+   * const domain = "thirdweb.com";
+   * // Generate a signed login payload for the connected wallet to authenticate with
+   * const loginPayload = await sdk.auth.login(domain);
+   * ```
+   */
+  public async login(
```
joaquim-verges, line 77: does the doc show this as @beta as well?

joaquim-verges (approved):
looks great - clean code.

Have some comments on naming / nit picks. Feel free to merge after addressing (also one test failed i think)

src/core/auth/wallet-authenticator.ts
adam-maj: Originally had the expiration times defaulted in the zod schema, but it was giving me an odd error where it would keep generating the same default time (which wasn't actually correct 5 minutes from now) and was causing errors until I removed it from schema, not sure what the issue was there but this seemed like an even trade.

adam-maj: As far as the timings go (5 minutes and the 5 hours one below), these are important security considerations, and it's okay to have them within a reasonable range. The signature is a one time use payload, ie you really only should need to use each signature once to prove your identity, and 5 minutes is a standard lifetime for a one time use/short-lived  token (along with 1 min, 2 mins, etc. - JWT access/refresh tokens usually fall in this range as well).

As far as the 5 hour one, that's another commonly used timing for session times/JWT token times (that don't rely on refresh tokens)

We definitely _never_ want to have an infinite expiration access token - it's a huge security concern because if someone steals your access token, they can hack your account for the remainder of the time that the token is valid, so adding expiration times mitigates this risk. If there is no expiration, the attacker can auth to your account forever, unless the server-side blocks the specific payload UUID or you switch account 😢 - so we definitely want to keep expiration defaulted on the JWT token.

src/core/auth/wallet-authenticator.ts
adam-maj: Responded above on this + the other timing question - TL;DR the timings are important, 5m/5h are standard/acceptable for what we're using them for

src/core/auth/wallet-authenticator.ts
adam-maj: Yep looks like all get marked as @beta in docs

PR #456, opened 2022-06-06
Add ERC721 claimable feature detection

adam-maj:


src/core/classes/drop-claim-conditions.ts
```diff
@@ -205,31 +209,37 @@ export class DropClaimConditions<
...
+            | DropERC20;
+          [validMerkleProof] = await wrapper.verifyClaimMerkleProof(
+            activeConditionIndex,
+            addressToCheck,
+            quantity,
+            proofs.proof,
+            proofs.maxClaimable,
+          );
+        } else {
+          const wrapper = this.contractWrapper.readContract as BaseDropERC721;
```
joaquim-verges, line 231: I think if you typeguard as `DropERC721 | DropERC20` above then this else will also be type guarded?

src/core/classes/erc-721-dropable.ts
```diff
@@ -0,0 +1,247 @@
...
+   *   },
+   *   {
+   *     startTime: publicSaleStartTime, // 24h after presale, start public sale
+   *     price: 0.08, // public sale price
+   *   }
+   * ]);
+   * await contract.claimConditions.set(claimConditions);
+   * ```
+   */
+  public claimConditions: DropClaimConditions<BaseDropERC721>;
```
joaquim-verges, line 57: so this is not optional right now, but it should be after the contract changes right? can you add a //TODO in there?

src/types/eips.ts
```diff
@@ -1,13 +1,16 @@
...
   IERC20Metadata,
   IERC721,
   IERC721Metadata,
+  LazyMintERC721,
 } from "contracts";
 
 export type BaseERC20 = IERC20 & IERC20Metadata;
 export type BaseERC721 = IERC721 & IERC721Metadata;
 export type BaseERC1155 = IERC1155 & IERC1155Metadata & IERC1155Supply;
+export type BaseDropERC721 = BaseERC721 & Drop & LazyMintERC721;
```
joaquim-verges, line 16: also a // TODO in here to update with new drop extension contract structure

src/core/classes/drop-claim-conditions.ts
adam-maj: Doesn't look like it type guards the else, guess it doesn't do set difference for the unioned types - it does work for the if part though at least

src/core/classes/erc-721-dropable.ts
adam-maj: This shouldn't be optional right - `IDrop` still depends on `IClaimCondition` which I think makes sense + `setClaimCondition` lives in `IDrop` so we couldn't really separate it out, are we making additionally interface changes besides whats on main? 

https://github.com/thirdweb-dev/contracts/blob/8e831c04dc38d761f7533ffe743c4d40e8f9a393/contracts/feature/interface/IDrop.sol#L51

adam-maj: I guess as per your notion doc, it does make sense to split them out in the future

PR #469, opened 2022-06-14
Add pack contract implementation

adam-maj:


src/schema/tokens/pack.ts
```diff
@@ -0,0 +1,97 @@
...
+  ERC721WrappableSchema,
+} from "./common/wrap";
+
+/**
+ * @internal
+ */
+const ERC20RewardSchema = ERC20WrappableSchema.omit({
+  quantity: true,
+}).extend({
+  quantityPerReward: PriceSchema,
```
joaquim-verges, line 16: wonder if it's really worth it to rename this. would help with code sharing if we didnt

src/schema/tokens/pack.ts
```diff
@@ -0,0 +1,97 @@
...
+  erc1155Rewards: z.array(ERC1155RewardSchema).default([]),
+});
+
+/**
+ * @internal
+ */
+export const PackRewardsOutputSchema = z.object({
+  erc20Rewards: z.array(ERC20RewardContentsSchema).default([]),
+  erc721Rewards: z.array(ERC721RewardContentsSchema).default([]),
+  erc1155Rewards: z.array(ERC1155RewardContentsSchema).default([]),
```
joaquim-verges, line 67: same for this - `erc20Contents` would make sense in both the pack and MW context right?

test/pack.test.ts
```diff
@@ -76,118 +82,103 @@ describe("Pack Contract", async () => {
...
-          amount: BigNumber.from(50),
+          quantityPerReward: 1,
+          totalRewards: 50,
         },
         {
+          contractAddress: bundleContract.getAddress(),
           tokenId: "2",
-          amount: BigNumber.from(50),
+          quantityPerReward: 1,
+          totalRewards: 50,
```
joaquim-verges, line 102: this does read kinda nice though

test/pack.test.ts
```diff
@@ -76,118 +82,103 @@ describe("Pack Contract", async () => {
...
         },
       ],
       metadata: {
```
joaquim-verges, line 105: should we call this "packMetadata" to be clear its for the pack?

src/contracts/pack.ts
```diff
@@ -121,448 +126,428 @@ export class Pack implements UpdateableNetwork {
...
+   * @param metadataWithRewards - the metadata and rewards to include in the pack
    *
    * @example
    * ```javascript
-   * // Data to create the pack
-   * const pack = {
-   *   // The address of the contract that holds the rewards you want to include
-   *   assetContract: "0x...",
-   *   // The metadata of the pack
+   * const packMetadata = {
```
joaquim-verges, line 292: i would call this `pack` and `metadata` -> `packMetadata`

joaquim-verges (commented):
overall looks good, minor things - mainly trying to avoid some code duplication if we can. Though the API reads pretty nice right now

src/contracts/pack.ts
```diff
@@ -49,17 +55,15 @@ import { ContractAnalytics } from "../core/classes/contract-analytics";
...
   static contractAbi = require("../../abis/Pack.json");
   /**
    * @internal
    */
-  static schema = PacksContractSchema;
-
-  private contractWrapper: ContractWrapper<PackContract>;
-  private storage: IStorage;
+  static schema = PackContractSchema;
+  // TODO: Change schema and deployment
```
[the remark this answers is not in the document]
adam-maj: Oops just forgot to remove the comment

src/contracts/pack.ts
```diff
@@ -121,448 +126,428 @@ export class Pack implements UpdateableNetwork {
...
+      const hasAllowance = await hasERC20Allowance(
+        this.contractWrapper,
+        erc20.contractAddress,
+        totalQuantity,
+      );
+      if (!hasAllowance) {
+        throw new Error(
+          `ERC20 token with contract address "${
+            erc20.contractAddress
+          }" does not have enough allowance to transfer.\n\nYou can set allowance to the multiwrap contract to transfer these tokens by running:\n\nawait sdk.getToken("${
```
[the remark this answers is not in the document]
adam-maj: There are some key differences because of the pack interface we want vs the one the contract implements, we need to do some multiplication/division in all these in the pack file that's different for each function in pack + different from multiwrap - which is why I opted to keep these all separate as verbose as it is, b/c handling all the casing seems messier and wanted to keep things explicit.

TL;DR, because contract uses total tokens + number of rewards (division), but we use tokens per reward + number of rewards (multiplication) we have to do some conversions between these interfaces which makes things different from MW and each other.

src/schema/tokens/pack.ts
adam-maj: I think quantityPerReward is an important distinction here - added it because I originally had just `quantity` as the interface, but when I was using it, it felt like quantity was the number of tokens I was going to wrap (which it isn't), so it felt confusing. What I had in mind was wouldn't want to have a user put something like:
```
quantity: 1
totalRewards: 50
```
expecting to only lose 1 token, and then lose 50, so wanted to make it explicit.

test/pack.test.ts
adam-maj: yep good idea, maybe should to something similar for MW

src/schema/tokens/pack.ts
adam-maj: Schema is a bit different for rewards vs MW because of the amounts stuff

src/contracts/pack.ts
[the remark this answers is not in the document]
joaquim-verges: i see - that makes sense.

src/contracts/pack.ts
```diff
@@ -121,448 +125,428 @@ export class Pack implements UpdateableNetwork {
...
-          this.contractWrapper.getProvider(),
-          rewardId.toString(),
-          this.storage,
-        ),
-    }));
+  /**
+   * Create Pack
+   * @remarks Create a new pack with the given metadata and rewards and mint it to the connected wallet.
+   *
+   * @param metadataWithRewards - the metadata and rewards to include in the pack
```
joaquim-verges, line 275: let's add a `See {@link Pack.createTo}` in the remarks

src/contracts/pack.ts
joaquim-verges: fix the example after the `metadata` rename?

PR #471, opened 2022-06-15
allow passing storage to fromSigner and fromPrivateKey - this was missed when first implementing those functions

jnsdls:


joaquim-verges (approved):
Nice catch

PR #472, opened 2022-06-16
Update code snippet example for Edition Drop Contract

OscarKevinXR:
In the Contract Dashboard the example provided is:
`await contract.airdrop(addresses, tokenId)`

But in the [documentation here](https://portal.thirdweb.com/typescript/sdk.Erc1155.airdrop) the parameter order is specified as
`airdrop(tokenId: BigNumberish, addresses: AirdropInput, data?: BytesLike): Promise<TransactionResult>;`

So you place `tokenId` before `addresses`, otherwise you get an `invalid BigNumber value` error upon executing the function.

This pull request is to update the code snippet to accurately reflect the contract function invocation.

### DIFF
--- a/src/core/classes/erc-1155.ts
+++ b/src/core/classes/erc-1155.ts
@@ -222,6 +222,8 @@ export class Erc1155<
    *
    * @example
    * ```javascript
+   * // The token ID of the NFT you want to airdrop
+   * const tokenId = "0";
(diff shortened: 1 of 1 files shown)


joaquim-verges:
Thank you for catching this! Will merge and generate docs which will then get reflected on the portal.

Thanks for contributing 🙏

PR #484, opened 2022-06-23
expose `sdk.getBalance(anyAddress)`

joaquim-verges:


joaquim-verges:
I thought about it, but for ERC20 we kinda overlap with `tokenContract.balanceOf()` so felt unneeded. Can revisit later though

PR #491, opened 2022-06-25
improve events queries (add txn data, ordering of events, BlockTags, better types, etc)

jnsdls:


src/core/classes/contract-events.ts
```diff
@@ -267,10 +284,15 @@ export class ContractEvents<TContract extends BaseContract> {
     event: EventFragment,
     args: ReadonlyArray<any>,
   ): ContractEvent {
-    const results: Record<string, any> = {};
+    const transaction = Object.fromEntries(
+      Object.entries(event).filter(
+        (a) => typeof a[1] !== "function" && a[0] !== "args",
```
joaquim-verges, line 289: if i remember correctly, the reason why we have 2 of these functions is because the format is slightly different between the 2. just double check.

also what's this `function` in there?

joaquim-verges (approved):

src/core/classes/drop-erc1155-history.ts
```diff
@@ -29,9 +29,17 @@ export class DropErc1155History {
...
+        ? e.data.tokenId.eq(tokenId)
+        : false,
     );
 
-    return Array.from(new Set(a.map((b) => b.data?.claimer)));
+    return Array.from(
+      new Set(
+        a
+          .filter((b) => typeof b.data?.claimer === "string")
+          .map((b) => b.data.claimer as string),
```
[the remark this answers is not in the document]
jnsdls: yes because I switched it to not return `any` instead `unknown` (because that's what we should be returning, users should have to check the types before they use stuff, we should never return `any` really)

src/core/classes/contract-events.ts
jnsdls: `event` has a bunch of FNs on it by default to parse more stuff, etc. I saw no reason to return those out with the data

src/types/events.ts
```diff
@@ -15,13 +17,15 @@ export interface UploadProgressEvent {
...
 
 /**
  * Filters for querying past events
  */
-export interface QueryAllEvents {
-  fromBlock?: number;
-  toBlock?: number;
+export interface EventQueryFilter {
+  fromBlock?: providers.BlockTag;
+  toBlock?: providers.BlockTag;
```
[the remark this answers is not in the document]
jnsdls: I mean I can make it `number | string` myself here (that's what it is)

jnsdls:
@joaquim-verges mind taking another look?

src/core/classes/contract-events.ts
```diff
@@ -73,15 +73,20 @@ export class ContractEvents<TContract extends BaseContract> {
       eventName as string,
     );
 
-    const wrappedListener = (...args: unknown[]) => {
-      // convert event info into nice object with named properties
-      const results = this.toContractEvent(event, args);
-      listener(results);
+    const address = this.contractWrapper.readContract.address;
+    const filter = { address, topics: [event.name] };
```
joaquim-verges, line 77: so we do need the address in filters? does that impact `getPastEvents` as well?

joaquim-verges (approved):
lg - ship it

PR #497, opened 2022-06-29
make getContract() resolve our built in contracts

joaquim-verges:
calling `getContract()` with an address of one of our known contracts now resolves (since we have the abi for it), still gives you a SmartContract instance but it will have all the extensions available. cc @jnsdls


jnsdls:
very nice!

PR #503, opened 2022-07-01
Rename IpfsUploader to PinataUploader

adam-maj:


src/core/classes/ipfs-storage.ts
```diff
@@ -297,7 +297,7 @@ export class IpfsStorage implements IStorage {
     signerAddress?: string,
   ): Promise<string> {
     // TODO move down to IStorageUpload
-    const token = await (this.uploader as IpfsUploader).getUploadToken(
+    const token = await (this.uploader as PinataUploader).getUploadToken(
```
adam-maj, line 300: Do we really need this function here? It's doing a lot of things that we shouldn't be doing - for one it's using an internal function of a single IPFS implementation in the top-level class (this won't work once we add more uploaders), and also its exposed publicly even though its marked as internal.

adam-maj: @joaquim-verges

PR #505, opened 2022-07-04
Update Signature Drop create batch with progress and improvement

nachoiacovino:


joaquim-verges (approved):
Thank you! Will try to consolidate these in the future so the logic is shared

PR #527, opened 2022-07-26
Add correct avax fuji chain name

nachoiacovino:


jnsdls (approved):
LGTM - ideally we should end up driving these names etc directly from SDK instead of having multiple sources of truth (dashboard and sdk are not guaranteed to be in sync right now)

PR #528, opened 2022-07-27
handle nextTokenIdToMint for BatchMintMetadata ERC721 contract extension

jnsdls:


jnsdls:
nah, this is not the way

PR #529, opened 2022-07-28
Rename PriceSchema to AmountSchema

nachoiacovino:
So we get a correct error message if they submit an invalid quantity


jnsdls (approved):
this will work fine but I dislike the duplication just for a different error message

src/schema/shared.ts
```diff
@@ -84,6 +84,13 @@ export const RawDateSchema = z.date().transform((i) => {
   return BigNumber.from(Math.floor(i.getTime() / 1000));
 });
 
+export const AirdropQuantitySchema = z
```
joaquim-verges, line 87: I think we already have an alias for Price that is "Amount" - can you use that instead?

nachoiacovino: not sure what you mean here, I added this because I realized the error message for airdrop when invalid quantity was wrong

joaquim-verges: lets just call this `AmountSchema` and use it for the `Amount` type

joaquim-verges (approved):

src/common/currency.ts
```diff
@@ -32,7 +32,7 @@ export async function normalizePriceValue(
   currencyAddress: string,
 ) {
   const metadata = await fetchCurrencyMetadata(provider, currencyAddress);
-  return utils.parseUnits(PriceSchema.parse(inputPrice), metadata.decimals);
+  return utils.parseUnits(AmountSchema.parse(inputPrice), metadata.decimals);
```
joaquim-verges, line 35: this was PriceSchema - why is it changed?

src/types/currency.ts
```diff
@@ -39,10 +39,10 @@ export type TokenHolderBalance = { holder: string; balance: CurrencyValue };
  * Represents a currency price already formatted. ie. "1" for 1 ether.
  * @public
  */
-export type Price = z.input<typeof PriceSchema>;
+export type Price = z.input<typeof AmountSchema>;
```
joaquim-verges, line 42: why did you change this one? this should not be changed

joaquim-verges (changes requested):

joaquim-verges (approved):
talked offline

PR #530, opened 2022-08-03
fix: do not minify compiled sources for Node.js

robertrossmann:
Minification makes debugging this project in its installed form practically impossible because

1. you cannot reasonably step through the code
2. variable names are unreadable

I understand the motivation to minify the codebase when targetting web browsers to reduce the overal size of the SDK but for Node.js file size is only a problem when it goes to the extremes.

I would like to also ask you, should you accept this PR, to also publish a non-minified version of the SDK to npm. 🙏

### DIFF
--- a/tsup.config.ts
+++ b/tsup.config.ts
@@ -12,7 +12,7 @@ export default defineConfig([
     sourcemap: true,
     // we'll just manually "clean" the dist dir before running this (to avoid potential race conditions)
     clean: false,
-    minify: true,
+    minify: false,
     platform: "node",
     replaceNodeEnv: true,
     // now required because not defaulted anymore


joaquim-verges:
Thanks for this @robertrossmann ! 

This change seems very reasonable to me. We need to run a couple of tests to check what happens on some browser frameworks that annoyingly load the node package instead of the browser one 😐 

Will let you know once it's done, we'll try to get to it this week. cc @jnsdls

robertrossmann:
Thanks @joaquim-verges ! ❤️ FYI merging and releasing this to npm would be a great help on my current project as it would allow me to step through the SDK's code when things don't work as I would expect. 😄 Also, this would allow me to apply hotfixes on my local installation of the SDK with `patch-package` and submit a proper pull request at a more convenient time than two days before product launch. 😂

joaquim-verges:
Merging this as well. Haven't had time to test much but it's easy to revert if reports come through.

PR #532, opened 2022-08-03
fix: restore compatibility with experimental built-in fetch in Node v18

robertrossmann:
As of Node.js v18, Node has a built-in, experimental support for global `fetch`. Unfortunately, this implementation is not compatible with `form-data` npm package and when an instance of this package's `FormData` class is provided as a body the native implementation resorts to serialising the form data by calling `.toString()` on it, causing the request's body to become the string literal `"[object FormData]"`.

We can work around that compatibility issue by explicitly using the form's headers and body as a buffer as the inputs to `fetch`.

### DIFF
--- a/src/core/classes/ipfs-storage.ts
+++ b/src/core/classes/ipfs-storage.ts
@@ -322,8 +322,9 @@ export class IpfsStorage implements IStorage {
       method: "POST",
       headers: {
         Authorization: `Bearer ${token}`,
+        ...formData.getHeaders(),
       },
-      body: formData as any,
+      body: formData.getBuffer(),
     });
     if (!res.ok) {
       throw new Error(`Failed to upload to IPFS [status code = ${res.status}]`);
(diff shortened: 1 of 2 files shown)


joaquim-verges:
Amazing, we haven't had much time to look into Node 18 support, so this helps a lot. Does everything else work for you with that fix? If so I can setup our automated tests to run on node 18 and have that running for all changes moving forward

robertrossmann:
@joaquim-verges well, this was a blocker for me on step one after installing the SDK so I honestly don't know how the rest of the SDK is looking on Node v18, this allows me to actually mint some tokens and that's a great start for me now. 😄 I will let you guys know if I discover any additional issues. 💪

joaquim-verges (approved):
approving to run the CI tests - still need to do some manual tests

joaquim-verges:
Couldn't get the node 18 tests to run because of a hardhat dependency that doesn't support it. But tested this on node 14/16 and browser and it seems fine. Merging. cc @jnsdls

PR #535, opened 2022-08-04
Add support for gasLimitOf function

adam-maj:


src/core/classes/gas-cost-estimator.ts
```diff
@@ -39,6 +39,31 @@ export class GasCostEstimator<TContract extends BaseContract> {
...
+   * const gasLimitOfClaim = await nftDrop?.estimator.gasLimitOf("claim", [
+   *   "0x...", // receiver
+   *   1, // quantity
+   *   "0x...", // currency
+   *   1, // price per token
+   *   [], // proofs
+   *   1, // proof max quantity per transaction
+   * ]);
+   * ```
+   * @returns the estimated gas limit in GWEI
```
joaquim-verges, line 57: comment is wrong

src/core/classes/gas-cost-estimator.ts
```diff
@@ -39,6 +39,31 @@ export class GasCostEstimator<TContract extends BaseContract> {
...
+   * @public
+   */
+  public async gasLimitOf(
+    fn: keyof TContract["functions"] | (string & {}),
+    args: Parameters<TContract["functions"][typeof fn]> | any[],
+  ): Promise<BigNumber> {
+    return this.contractWrapper.estimateGas(fn, args);
```
joaquim-verges, line 64: can you write a quick test for this for sanity? can just check that gasLimit > 0

src/core/classes/gas-cost-estimator.ts
adam-maj: What about it were you referring to - something besides `wei` instead of `gwei`?

adam-maj: ah yes - just gas units

joaquim-verges (approved):

PR #536, opened 2022-08-05
Introducing `TransactionTask` builder pattern for fine grained transaction control

joaquim-verges:
Beginning with just the drop claim functions, to give more granularity for the transaction, introduced a new API:

```javascript
const tx = dropContract.getClaimTransaction(receiver, quantity);
// you can override stuff
tx.overrideGasLimit(...)
tx.overrideNonce(...)
// you can run estimations
tx.estimateGasCostInEther(...)
// you can encode tx data
tx.encode()
// you can submit without waiting for confirmations
tx.submit()
// you can execute (what we do today)
tx.execute()
```

Eventually we'll re-use this `TransactionTask` object for other transactions, exposing that same fine grained control


src/core/classes/TransactionTask.ts
```diff
@@ -0,0 +1,124 @@
...
+    return await this.estimator.gasLimitOf(this.functionName, this.args);
+  }
+
+  public async estimateGasCostInEther(): Promise<string> {
+    return await this.estimator.gasCostOf(this.functionName, this.args);
+  }
+
+  // ////////////// Actions ////////////////
+
+  public async encode(): Promise<string> {
```
adam-maj, line 102: Can we get methods off this to get all parts of tx data - this `encode` function gives you the `data` param, but would be nice to get everything including `nonce`, `value`, `gasLimit`, etc. all in one structure from a function (similar to how fractal needs it). This would actually let us create this generic contract encoder interface for every function call.

adam-maj: Here's an example of what the structure looks like: https://buildthirdweb.slack.com/archives/C03GRV58Y13/p1658382917
Screen Shot 2022-08-05 at 12 45 04 PM

And implementation in Go: https://buildthirdweb.slack.com/archives/C03GRV58Y13/p165844326355217
Screen Shot 2022-08-05 at 12 45 33 PM

adam-maj: But maybe most convenient is just to have a single function that returns the entire tx object, however ethers expects it, so people can do signing themselves as well (or at least for now to access the data).

src/core/classes/TransactionTask.ts
```diff
@@ -0,0 +1,124 @@
...
+    return this.encoder.encode(this.functionName, this.args);
+  }
+
+  public async submit(): Promise<ContractTransaction> {
```
adam-maj, line 106: Should we add docs on this interface as well (like on `submit` vs `execute`)

src/core/classes/TransactionTask.ts
```diff
@@ -0,0 +1,124 @@
...
+import { GasCostEstimator } from "./gas-cost-estimator";
+import { TransactionResult } from "../types";
+
+class TransactionTaskBuilder {
+  public contractWrapper: ContractWrapper<any>;
+  public functionName: string;
+  public args: any[] = [];
+  public overrides: CallOverrides | undefined;
+
+  constructor(contractWrapper: ContractWrapper<any>, functionName: string) {
```
adam-maj, line 13: Would it be beneficial to use/at least support the argument object pattern which tends to be common in JS vs the builder pattern. We had builder before with claim conditions and moved away from it - these two aren't as complex (this comment applies to both `TransactionTaskBuilder` and `TransactionTask`) but could be more convenient

src/contracts/signature-drop.ts
```diff
@@ -463,6 +464,38 @@ export class SignatureDrop extends Erc721<SignatureDropContract> {
...
+  public async getClaimTransaction(
+    destinationAddress: string,
+    quantity: BigNumberish,
+    checkERC20Allowance = true, // TODO split up allowance checks
+  ): Promise<TransactionTask> {
+    const claimVerification = await this.prepareClaim(
+      quantity,
+      checkERC20Allowance,
+    );
+    return TransactionTask.builder(this.contractWrapper, "claim")
```
adam-maj, line 483: **nitpick** would be more concise with object params pattern instead of factory/builder and (imo but maybe wrong) more idiomatic for JS

adam-maj: Also, instead of passing `this.contractWrapper` here, could we add a function called `buildTransaction` to `contractWrapper` that lets us do `this.contractWrapper.buildTransaction` with the exact same API we've been using before for `sendTransaction`

src/core/classes/TransactionTask.ts
joaquim-verges: yeah that's interesting basically make this one `encodeFunctionData()` and add another `encodeTransaction()` or something like that

src/core/classes/TransactionTask.ts
joaquim-verges: yep definitely, adding it now

src/core/classes/TransactionTask.ts
joaquim-verges: I see what you mean - I kinda gravitated towards a builder pattern mainly coz im used to it. Might not be the JS way. I'll play with an object as argument, prob less code too

PR #538, opened 2022-08-09
Add support for erc1155 lazy mint

adam-maj:


joaquim-verges (approved):
very nice!

jadbox:
@adam-maj Isn't there already support for 1155 lazy mint?

PR #540, opened 2022-08-10
Add support for burnable to ERC20, ERC721, and ERC1155

adam-maj:


src/contracts/nft-drop.ts
```diff
@@ -554,14 +556,12 @@ export class NFTDrop extends Erc721<DropERC721> {
...
-   * const result = await contract.burn(tokenId);
+   * const result = await contract.burnFromSelf(tokenId);
    * ```
    *
    */
-  public async burn(tokenId: BigNumberish): Promise<TransactionResult> {
-    return {
-      receipt: await this.contractWrapper.sendTransaction("burn", [tokenId]),
-    };
+  public async burnFromSelf(tokenId: BigNumberish): Promise<TransactionResult> {
```
joaquim-verges, line 563: let's call this `burnToken(tokenId)` reads much nicer

src/core/classes/erc-721-burnable.ts
```diff
@@ -0,0 +1,35 @@
...
+   *
+   * @example
+   * ```javascript
+   * // The token ID of the NFT you want to burn
+   * const tokenId = 0;
+   *
+   * await contract.nft.burn.fromSelf(tokenId);
+   * ```
+   */
+  public async fromSelf(tokenId: BigNumberish): Promise<TransactionResult> {
```
joaquim-verges, line 30: let's also call this `token(tokenId)`

so it goes `contract.nft.burn.token(tokenId)`

src/contracts/token.ts
```diff
@@ -270,11 +268,6 @@ export class Token extends Erc20<TokenERC20> {
     holder: string,
```
joaquim-verges, line 268: let's get rid of this function, it's just confusing and has almost no value

joaquim-verges (changes requested):

src/contracts/token.ts
adam-maj: Feels odd to remove this (assuming you're referring to `burnFrom` because its the only function on the token interface that calls the `burnFrom` function on the contract: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/ad67d48f0914a15a9350ef42672322058b48d5b1/contracts/token/ERC20/extensions/ERC20BurnableUpgradeable.sol#L41

Also, it's existed on our token contract for a long time now. We should probably have contract interface for the functions on the contract right? Otherwise just shouldn't be on the contract if it really has no use (but since OZ has it, maybe people use it for something).

adam-maj: Will delete anyway for now

src/contracts/token-drop.ts
```diff
@@ -256,43 +259,11 @@ export class TokenDrop extends Erc20<DropERC20> {
...
-   * // Address of the wallet sending the tokens
-   * const holderAddress = "{{wallet_address}}";
-   *
-   * // The amount of this token you want to burn
-   * const amount = 1.2;
-   *
-   * await contract.burnFrom(holderAddress, amount);
-   * ```
-   */
-  public async burnFrom(
```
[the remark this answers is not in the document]
adam-maj: There's only two functions in the extension though - `burn` and `burnFrom`. Isn't it odd to only support one of them if someone intentionally implemented?

joaquim-verges (approved):
shipit

PR #547, opened 2022-08-16
Add support for ERC20Droppable

adam-maj:


src/core/classes/erc-20-droppable.ts
```diff
@@ -0,0 +1,129 @@
...
+   * const receipt = tx.receipt; // the transaction receipt
+   * ```
+   *
+   * @param destinationAddress - Address you want to send the token to
+   * @param amount - Quantity of the tokens you want to claim
+   * @param checkERC20Allowance - Optional, check if the wallet has enough ERC20 allowance to claim the tokens, and if not, approve the transfer
+   *
+   * @returns - The transaction receipt
+   */
+  public async claimTo(
```
joaquim-verges, line 94: can you call this inside `TokenDrop.claimTo()` so we remove some duplication?

joaquim-verges (approved):
just one deduplication comment, otherwise looks good to go

joaquim-verges (approved):

PR #561, opened 2022-08-24
make input/output versions of the release metadata schemas

joaquim-verges:


jnsdls:
ported here: https://github.com/thirdweb-dev/js/pull/13

